PR #50, opened 2025-12-25
Pad 262 restrict same password as old

shavtvalishvili:
[262](https://github.com/noxtton/pearpass-app-desktop/issues/262)

### Requirements

- The system should not allow using the same value for the old and new password.
- A validation message should be shown, “New password must be different from the current password.”.

### Changes

- Use password change validation utility function from pearpass--utils-password-check

### Screenshots/Recordings

- Desktop (MacOS 26.0.1)
https://github.com/user-attachments/assets/bfeee664-a35f-4f81-a9ff-9ee574224412

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -402,7 +402,7 @@
     },
     "node_modules/@babel/plugin-bugfix-firefox-class-in-computed-class-key": {
       "version": "7.28.5",
-      "dev": true,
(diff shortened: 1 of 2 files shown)


eugeneglova (approved):

torkinos:
I'll be taking over this update, there are some other changes needed

PR #77, opened 2026-01-14
Prevent duplicate recording of failed master password attempts

ndolidzee:
### Requirements
After each failed login attempt, the remaining login attempts should be reduced by 1

### Changes
Removed duplicate recording of failed login attempts

### Screenshots/Recordings

https://github.com/user-attachments/assets/b409725c-4353-444e-a053-2dd07e12a5ae

### DIFF
--- a/src/services/handlers/EncryptionHandlers.js
+++ b/src/services/handlers/EncryptionHandlers.js
@@ -64,15 +64,11 @@ export class EncryptionHandlers {
       hashedPassword: params.hashedPassword
     })
 
-    if (!result) {
-      logger.info('ENCRYPTION-HANDLER', `Decryption failed, recording attempt`)
-      await this.client.recordFailedMasterPassword()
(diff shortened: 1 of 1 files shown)


src/services/handlers/EncryptionHandlers.js
```diff
@@ -64,15 +64,11 @@ export class EncryptionHandlers {
       hashedPassword: params.hashedPassword
     })
 
-    if (!result) {
-      logger.info('ENCRYPTION-HANDLER', `Decryption failed, recording attempt`)
-      await this.client.recordFailedMasterPassword()
```
denislupookov: Is there another way to fix it without removing recordFailedMasterPassword here ?
I had PR to fix Master Password Lockout Bypass via IPC - https://github.com/tetherto/pearpass-app-desktop/pull/61

PR #88, opened 2026-01-15
Native messaging handshake vulnerability

shavtvalishvili:
### Changes

- Require the extension to provide its Ed25519 public key during pairing, store it and use it to gate future handshakes.
- Remove the separate isPairingApproved flag, the presence of a stored client public key now indicates pairing is complete.
- Include the pinned client public key in the handshake transcript so both parties must agree on which extension identity is being used.
- Reject pairing attempts from a different extension if one is already paired, require explicit reset first.
- Add a 1-hour session TTL; expired sessions are automatically deleted on access, forcing a transparent re-handshake.
- Keep a handshake transcript and clientVerified flag on each session; only mark verified after successful client signature check.
- Return clear NotPaired / ClientNotVerified / ClientAlreadyPaired errors instead of silently proceeding.
- Updated unit tests for security handlers and session management.

### Testing Notes

- Tested agains a local script trying to access IPC via existing implemented protocol and handshake commands which can no longer get vault data as it could before these changes.
- Tested extension pairing and repairing, authentication.

### dependencies

- https://github.com/tetherto/pearpass-app-browser-extension/pull/32


src/services/handlers/SecurityHandlers.js
```diff
@@ -72,9 +106,38 @@ export class SecurityHandlers {
...
+
+    // Load pinned client identity
+    const clientPubB64 = await getClientIdentityPublicKey(this.client)
+    if (!clientPubB64) {
+      throw new Error('ClientNotPaired: No client identity registered')
+    }
+
+    const clientPubBytes = new Uint8Array(Buffer.from(clientPubB64, 'base64'))
+    const sigBytes = new Uint8Array(Buffer.from(clientSigB64, 'base64'))
+
```
alvaropaco: Add explicit length checks (Ed25519 = 32-byte pubkey, 64-byte signature):
```javascript
if (clientPubBytes.length !== 32) throw new Error('InvalidClientPublicKey')
if (sigBytes.length !== 64) throw new Error('InvalidClientSignature')
if (!session.transcript || session.transcript.length === 0) throw new Error('InvalidTranscript')
```

src/services/security/appIdentity.js
```diff
@@ -44,6 +46,29 @@ const toBase64 = (bytes) => Buffer.from(bytes).toString('base64')
...
+  )
+
+  if (!pairingSecretB64) {
+    const secretBytes = new Uint8Array(32)
+    sodium.randombytes_buf(secretBytes)
+    pairingSecretB64 = Buffer.from(secretBytes).toString('base64')
+
+    await client
+      .encryptionAdd(ENC_KEY_PAIRING_SECRET, pairingSecretB64)
+      .catch(() => {})
```
alvaropaco: Should ignore failures here? Maybe swallowing encryptionAdd() failure creates “ephemeral secret” pairing behavior.The function still returns pairingSecretB64, but it was not persisted. On next run you generate a different secret → pairing breaks unpredictably, and you may get subtle auth failures that look like “handshake issues”.
That makes sense?

src/services/security/appIdentity.js
```diff
@@ -44,6 +46,29 @@ const toBase64 = (bytes) => Buffer.from(bytes).toString('base64')
...
+ */
+const getOrCreatePairingSecret = async (client) => {
+  let pairingSecretB64 = normalizeEncryptionGet(
+    await client.encryptionGet(ENC_KEY_PAIRING_SECRET).catch(() => null)
+  )
+
+  if (!pairingSecretB64) {
+    const secretBytes = new Uint8Array(32)
+    sodium.randombytes_buf(secretBytes)
+    pairingSecretB64 = Buffer.from(secretBytes).toString('base64')
```
alvaropaco:
```suggestion
    pairingSecretB64 = Buffer.from(secretBytes).toString('base64')
    
    if (pairingSecretB64) {
  const bytes = Buffer.from(pairingSecretB64, 'base64')
  if (bytes.length !== 32) {
    throw new Error('InvalidPairingSecret')
  }
}
```

src/services/security/appIdentity.js
```diff
@@ -208,18 +240,33 @@ export const getOrCreateIdentity = async (client) => {
...
-export const getPairingCode = (ed25519PublicKeyB64) => {
-  const fingerprint = getFingerprint(ed25519PublicKeyB64)
-  // Use first 5 bytes -> numeric short code (e.g., 6 digits)
-  const firstBytes = Buffer.from(fingerprint.slice(0, 8), 'hex') // 4 bytes
-  const num = firstBytes.readUInt32BE(0)
+export const getPairingCode = (ed25519PublicKeyB64, pairingSecretB64) => {
+  const secret = fromBase64(pairingSecretB64)
+  const publicKey = fromBase64(ed25519PublicKeyB64)
+
+  // Compute H = SHA-256(secret || publicKey)
```
alvaropaco: This is fine cryptographically, but without a fixed tag/version, the same construction could accidentally be reused in another context and create cross-protocol coupling.
Maybe could be good do something like:
```javscript
const tag = Buffer.from('pearpass/pairingcode/v1', 'utf8')
const input = new Uint8Array(tag.length + secret.length + publicKey.length)
input.set(tag, 0)
input.set(secret, tag.length)
input.set(publicKey, tag.length + secret.length)
```

PR #112, opened 2026-01-28
CopyButton to input fields.

harrisaar-noxtton:
# Before

1. Clicking a field automatically copied its value to the clipboard.
2. Clicking a web field opened the URL in the browser instead of copying it.
3. If "Copy to clipboard" was disabled in settings, the copy action did not trigger.
4. TextArea did not have any copy functionality

# Now

1. Each field now features a dedicated copy icon.
2. Clicking the copy icon copies the field value to the clipboard.
3. The copy icon is hidden while in edit mode.
4. If "Copy to clipboard" is disabled, copy icons are not rendered.
5. Web fields with URLs display an "external link" icon; clicking it opens the URL in the browser.
6. Implemented similar copy feature to TextArea input.

# Code Changes

1. We created a dedicated `CopyButton` component, which now manages its own copy logic and visibility settings.
2. The `CopyButton` is plugged in using the `additionalItems` prop that was already implemented in `InputField`.
3. I converted a few files to TypeScript, though I tried to avoid touching too many files at once.
4. Created a `WebsiteButton` which, like the copy button, can be plugged into fields using `additionalItems`.

[demo copy icons.webm](https://github.com/user-attachments/assets/ad07de0c-dafe-466b-bd1d-53809c08cae8)

### DIFF
--- a/src/containers/CustomFields/index.tsx
+++ b/src/containers/CustomFields/index.tsx
@@ -8,7 +8,7 @@ import { CopyButton } from '../../components/CopyButton'
 interface CustomField {
   id: string
   type: 'note'
-  props: Record<string, any>
+  props: Record<string, unknown>
 }
 
 interface CustomFieldsProps {
@@ -16,7 +16,7 @@ interface CustomFieldsProps {
     name: string;
     value: string;
     error?: string;
-    onChange: (e: any) => void;
+    onChange: (e: unknown) => void;
   }
   customFields?: CustomField[]
   onClick?: () => void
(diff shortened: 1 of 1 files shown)


src/components/CopyButton/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+
+interface CopyButtonProps {
+  value?: string
+  testId?: string
+}
+
+const CopyButton = ({ value, testId }: CopyButtonProps): React.ReactElement => {
+
+  const { i18n } = useLingui()
```
torkinos: @harrisaar-noxtton I would suggest to use `useTranslation` hook instead

PR #134, opened 2026-02-02
Enhance clipboard management

torkinos:
### Changes

- Fix new additional Pear app running in Pear v2
- Fix not clearing issue when app is closed

### Screenshots/Recordings

https://github.com/user-attachments/assets/cc98e078-94f5-4b2b-942e-52cd4a8fc51e

---
  - https://app.asana.com/0/0/1212392291502304

### DIFF
--- a/subProcess/clipboardCleanup.js
+++ b/subProcess/clipboardCleanup.js
@@ -201,29 +201,32 @@ export function getClipboardContent() {
   })
 }
 
-;(async () => {
-  logger.log('Clipboard cleanup worker started')
+// Only run worker code when executed as a Pear subprocess, not when imported for testing
+// Check for Pear.exit to ensure we're in the actual Pear runtime (not Jest)
+if (typeof Pear !== 'undefined' && typeof Pear.exit === 'function') {
+  ;(async () => {
+    logger.log('Clipboard cleanup worker started')
 
-  // Get the text to monitor from command line args (passed by useCopyToClipboard)
-  const copiedValue = await getClipboardContent()
+    // Get the text to monitor from command line args (passed by useCopyToClipboard)
+    const copiedValue = await getClipboardContent()
 
-  // Only run worker code if we have args (running as a worker, not imported for testing)
-  // Convert timeout from ms to seconds
-  const timeoutSeconds = Math.ceil(CLIPBOARD_CLEAR_TIMEOUT / 1000)
+    // Only run worker code if we have args (running as a worker, not imported for testing)
+    // Convert timeout from ms to seconds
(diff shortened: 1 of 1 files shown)


denislupookov (approved):

src/hooks/useCopyToClipboard.js
```diff
@@ -59,14 +59,30 @@ export const useCopyToClipboard = ({ onCopy } = {}) => {
...
-          Pear.config.key ? CLEAR_CLIPBOARD_PATH_PROD : CLEAR_CLIPBOARD_PATH_DEV
-        )
-        pipeRef.current.write(text)
+        // Spawn worker with text as command-line argument
+        // This avoids the pipe closing issue that caused premature worker exit
+        const workerPath = Pear.config.key
+          ? CLEAR_CLIPBOARD_PATH_PROD
+          : CLEAR_CLIPBOARD_PATH_DEV
+
+        pipeRef.current = pearRun(workerPath, [text])
```
shavtvalishvili: When you call pearRun(workerPath, [text]), the text is appended to spawn's command line arguments - these are visible via ps aux or task manager as I know, maybe this can be a concern?

shavtvalishvili: Can't we still use pipes to avoid this?

subProcess/clipboardCleanup.js
```diff
@@ -0,0 +1,295 @@
...
+      // The script file approach avoids command line escaping issues.
+      const base64Value = Buffer.from(copiedValue).toString('base64')
+      const scriptId = Date.now()
+      const tempDir = os.tmpdir()
+      const scriptPath = path.join(tempDir, `pearpass_clip_${scriptId}.ps1`)
+
+      // PowerShell script content - sleeps, checks clipboard, clears if unchanged, deletes itself
+      const scriptContent = `
+Start-Sleep -Seconds ${timeoutSeconds}
+$v = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String('${base64Value}'))
```
shavtvalishvili: So this way the password is written on disk to a temporary file. Just to clarify if we're okay with this, though don't know what would be a working solution without this.

torkinos: I agree its a reasonable concern but also not sure with with the fix because of certain limitations.
But its not a critical vulnerability because this is an user scoped temp directory, also during this time the item is also accessible via clipboard.
I will create a separate ticket, to get back to this in future

shavtvalishvili: Ok, I agree.

PR #139, opened 2026-02-05
Fixed error An item is not created after unsuccessful Vault importing

denislupookov:
### Requirements

An item is created after unsuccessful Vault importing

### Changes

Bump packages and added loading on vault

### Demo

https://github.com/user-attachments/assets/140ccba6-037b-4821-b8c6-21211b36d707

---
  - https://app.asana.com/0/0/1212851114040885


torkinos (approved):

torkinos:
@denislupookov tests failing

PR #416, opened 2026-04-15
fix: flatpak clipboard issues

alvaropaco:
### Requirements

  - Flatpak build of PearPass must launch to a rendered UI on both X11 and      
  Wayland host sessions (no blank window).
  - Clipboard auto-clear must work inside the Flatpak sandbox in both scenarios:
    a. After the configured delay (default 30s) while the app is open.          
    b. Immediately when the user closes the app before the delay elapses.       
                                                                                
  ### Changes                                                                       
                                                                                
  - flatpak/com.pears.pass.yaml — switched the wrapper from --disable-gpu to    
  --use-gl=angle --use-angle=swiftshader. Forcing SwiftShader via ANGLE fixes
  the blank window that --disable-gpu alone produced in the GNOME 49 runtime    
  (software rendering fell through to a non-functional GL path).
  - electron/linuxWaylandClipboard.cjs — isWaylandSession() now returns false
  when GDK_BACKEND=x11 (or ELECTRON_OZONE_PLATFORM_HINT=x11). Inside the Flatpak
   the wrapper pins Electron to X11/XWayland, but the host env still advertises
  Wayland, which was routing clipboard reads to wl-paste against a socket that  
  isn't granted to the sandbox.                             
  - electron/clipboardCleanup.cjs
    - In X11-mode Flatpak (and in non-Flatpak runs), the scheduled-cleanup tick 
  and the on-quit clear now use Electron's own clipboard.readText() /           
  clipboard.clear() directly instead of shelling out. wl-paste/wl-copy were     
  failing for lack of the wayland socket and xsel/xclip aren't shipped by the   
  GNOME runtime, so the content-match check was comparing the live secret to
  null and skipping the clear entirely.
    - On quit, after clipboard.clear() we also invoke the bundled xsel via
  spawnSync (extracting the binary to $XDG_RUNTIME_DIR and chmod'ing it         
  executable). Electron's clear returns before the X SelectionOwner change is
  flushed, which on GNOME let the clipboard manager resurrect the old secret    
  from the dying owner; a synchronous xsel --clipboard --clear forces a real X
  round-trip before the process exits.
    - Removed the legacy useHostClipboard = isFlatpakRuntime() branch that
  forced every Flatpak run down the shell-out path.                             
  - electron/main.cjs — wires flushPendingClipboardCleanups into the before-quit
   cleanup path so pending auto-clear timers are flushed synchronously (and now 
  invoke the xsel fallback) before the process tears down.  
  - electron/clipboardCleanup.test.js — added coverage for the new              
  preferElectronClipboard branch, the on-quit flush path, and the xsel fallback.
   
 ### Testing Notes                                                                 
                                                            
  Built and ran the Flatpak on an Ubuntu/GNOME host in a QEMU VM:               
   
  1. flatpak run --user com.pears.pass — app launches and renders the UI (no    
  blank window).                                            
  2. Open a vault item → copy a field → wait 30s → paste into the app's search  
  field → clipboard is empty, nothing pastes.                                   
  3. Open a vault item → copy a field → immediately close the app → paste into
  another application (e.g. a text editor) → clipboard is empty, nothing pastes.
  4. Ran npm test — existing clipboard tests plus new cases pass.
                                                                                
  Things reviewers should pay attention to                  
                                                                                
  - The preferElectronClipboard condition keys off GDK_BACKEND /                
  ELECTRON_OZONE_PLATFORM_HINT. If we ever add a Wayland-socket Flatpak variant,
   we'll want to drop GDK_BACKEND=x11 from the wrapper or this branch will still
   route to Electron's X11 clipboard.                       
  - The bundled xsel is extracted to $XDG_RUNTIME_DIR (falling back to
  os.tmpdir()). Confirm this is acceptable for packagers — /tmp is often noexec 
  inside a Flatpak sandbox, which is why we avoid it.
  - --use-gl=angle --use-angle=swiftshader alongside --disable-gpu is unusual;  
  the combination is what actually renders on the target runtime, but if you    
  have a cleaner GPU-flag set that also works, happy to adopt it.
  - clipboard.writeText('') right after clipboard.clear() on quit is intentional
   — it's there to take X selection ownership with empty content before xsel    
  runs.
                                                                                
  ### Screenshots/Recordings                                    

  N/A — behavior verified via terminal logs and manual paste tests; can attach a
   screen recording on request.
                                                                                
### dependencies                                              

  None.


boris91 (approved):

giorgikh93:
@alvaropaco please attach screen recording.

PR #419, opened 2026-04-16
enhance ci to include format and type check

kapildev421:
### Requirements
Enhance CI to include
1. Format check
2. Type check

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,3 +28,14 @@ jobs:
 
       - name: Run ESLint
         run: npm run lint
+
+      
+      # Step : Run the Format Check (e.g., Prettier)
+      # If the code is formatted poorly, the pipeline stops here and fails.
+      - name: Check Formatting
+        run: npm run format:check
+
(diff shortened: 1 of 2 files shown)


boris91 (dismissed):

.github/workflows/lint.yml
```diff
@@ -28,3 +28,14 @@ jobs:
 
       - name: Run ESLint
         run: npm run lint
+
+      
+      # Step : Run the Format Check (e.g., Prettier)
+      # If the code is formatted poorly, the pipeline stops here and fails.
+      - name: Check Formatting
+        run: npm run format:check
```
hakaari, line 36: I don't think we have those scripts in package.json, so we need to define them there as well.

hakaari (changes requested):

.github/workflows/lint.yml
kapildev421: I added  but please make sure it is correct as i not expert developer

hakaari (approved):

torkinos (approved):

PR #442, opened 2026-04-22
Redesign of Languages & App version screens

alvaropaco:
Language Figma link: https://www.figma.com/design/c1uWBqMmX4Daw0uiVjfJZW/PearPass-%F0%9F%8D%90?node-id=12430-95795&m=dev

App Version Figma link: https://www.figma.com/design/c1uWBqMmX4Daw0uiVjfJZW/PearPass-%F0%9F%8D%90?node-id=12443-12214&m=dev

https://github.com/user-attachments/assets/40b64680-ce7e-430e-a57d-06ce07f7a860

### DIFF
--- a/src/pages/SettingsViewV2/content/AppVersionContent/index.tsx
+++ b/src/pages/SettingsViewV2/content/AppVersionContent/index.tsx
@@ -1,0 +1,133 @@
+import React, { useEffect, useState } from 'react'
+
+import {
+  Link,
+  PageHeader,
+  Text,
+  useTheme
+} from '@tetherto/pearpass-lib-ui-kit'
+import { PRIVACY_POLICY, TERMS_OF_USE } from '@tetherto/pearpass-lib-constants'
+
+import { useToast } from '../../../../context/ToastContext'
+import { useTranslation } from '../../../../hooks/useTranslation'
+import { logger } from '../../../../utils/logger'
+import { createStyles } from './styles'
+
+const WEBSITE_URL = 'https://pass.pears.com'
+
+const TEST_IDS = {
+  root: 'settings-app-version',
+  fieldVersion: 'settings-app-version-field',
+  checkForUpdates: 'settings-app-version-check-for-updates'
+} as const
+
+export const AppVersionContent = (): React.ReactElement => {
+  const { t } = useTranslation()
+  const { theme } = useTheme()
+  const { setToast } = useToast()
+  const styles = createStyles(theme.colors)
+
+  const [currentVersion, setCurrentVersion] = useState('')
+  const [isChecking, setIsChecking] = useState(false)
+
+  useEffect(() => {
+    const electronAPI = window.electronAPI
+    if (!electronAPI || typeof electronAPI.getConfig !== 'function') {
+      return
+    }
+
+    electronAPI
+      .getConfig()
+      .then((cfg) => {
+        if (cfg && typeof cfg.version === 'string') {
+          setCurrentVersion(cfg.version)
+        }
+      })
+      .catch((error) =>
+        logger.error(
+          'AppVersionContent',
+          'Error getting runtime config:',
+          error
+        )
+      )
+  }, [])
+
+  const handleCheckForUpdates = async () => {
+    const electronAPI = window.electronAPI
+    if (isChecking || !electronAPI?.checkUpdated) {
+      return
+    }
+
+    try {
+      setIsChecking(true)
+      const hasUpdate = await electronAPI.checkUpdated()
+
+      setToast({
+        message: hasUpdate
+          ? t('Update available. Restart the app to apply it.')
+          : t('You are on the latest version.')
+      })
+    } catch (error) {
+      logger.error(
+        'AppVersionContent',
+        'Error checking for updates:',
+        error
+      )
+      setToast({
+        message: t('Could not check for updates. Please try again.')
+      })
+    } finally {
+      setIsChecking(false)
+    }
+  }
+
+  return (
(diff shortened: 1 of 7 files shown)


boris91 (dismissed):

src/pages/SettingsViewV2/content/AppVersionContent/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <div style={styles.fieldContainer} data-testid={TEST_IDS.fieldVersion}>
+        <Text variant="bodyEmphasized" color={theme.colors.colorTextPrimary}>
+          {t('App version')}
+        </Text>
+        <Text variant="body" color={theme.colors.colorPrimary}>
+          {currentVersion}
+        </Text>
+      </div>
+
+      <button
```
giorgikh93: please use Button from ui-kit

src/pages/SettingsViewV2/content/LanguageContent/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+            {t('Select the language used throughout PearPass.')}
+          </Text>
+        </div>
+
+        <Dropdown
+          open={isOpen}
+          onOpenChange={setIsOpen}
+          testID={TEST_IDS.dropdown}
+          trigger={
+            <button
```
giorgikh93: Please use Button from ui-kit

src/pages/SettingsViewV2/content/AppVersionContent/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+  useTheme
+} from '@tetherto/pearpass-lib-ui-kit'
+import { PRIVACY_POLICY, TERMS_OF_USE } from '@tetherto/pearpass-lib-constants'
+
+import { useToast } from '../../../../context/ToastContext'
+import { useTranslation } from '../../../../hooks/useTranslation'
+import { logger } from '../../../../utils/logger'
+import { createStyles } from './styles'
+
+const WEBSITE_URL = 'https://pass.pears.com'
```
giorgikh93: please import PEARPASS_WEBSITE from pearpass-lib-constants

src/pages/SettingsViewV2/content/AppVersionContent/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+      .catch((error) =>
+        logger.error(
+          'AppVersionContent',
+          'Error getting runtime config:',
+          error
+        )
+      )
+  }, [])
+
+  const handleCheckForUpdates = async () => {
```
giorgikh93: This flow doesn't make sense for me because we already have a [custom hook](https://github.com/tetherto/pearpass-app-desktop/blob/main/src/hooks/usePearUpdate.js#L38) which is called in the [App.js](https://github.com/tetherto/pearpass-app-desktop/blob/main/src/app/App/index.js#L23) and if user opens the app and if update is available, update modal will be displayed instantly with countdown timer and updates automatically. I think it's better to discuss this with @costin-busioc - specifically how the flow should look like and what UX benefits we have by having this button. Maybe it's not needed.

giorgikh93 (changes requested):

src/pages/SettingsViewV2/content/AppVersionContent/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <PageHeader
+        as="h1"
+        title={t('App Version')}
+        subtitle={
+          <>
+            {t('Here you can find all the info about your app.')}
+            <br />
+            {t('Check here to see the')}{' '}
+            {/* @ts-ignore - plain CSS object */}
+            <Link href={TERMS_OF_USE} isExternal style={styles.descriptionLink}>
```
denislupookov, line 40: are u sure we need style for link ?

src/pages/SettingsViewV2/content/AppVersionContent/index.tsx
alvaropaco: I agreed! Just following the figma's desing.Could we let to remove it if we decide that it's not necessary in a another task? Just to avoid delays.

src/pages/SettingsViewV2/content/AppVersionContent/index.tsx
alvaropaco: @denislupookov Yes if we want to follow Figma's styles as is. The Link style is necessary because the current text variant inherits a 400 weight, while the design spec requires 500.  The override ensures the link appears emphasized without affecting size or default link behavior, though it could alternatively be handled via a more verbose variant-based approach.

giorgikh93 (approved):
LGTM thanks, @denislupookov please check

boris91 (dismissed):

src/pages/SettingsViewV2/content/AppVersionContent/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  root: 'settings-app-version',
+  fieldVersion: 'settings-app-version-field'
+} as const
+
+export const AppVersionContent = (): React.ReactElement => {
+  const { t } = useTranslation()
+  const { theme } = useTheme()
+  const styles = createStyles(theme.colors)
+
+  const [currentVersion, setCurrentVersion] = useState('')
```
giorgikh93: @alvaropaco could you please lift the state up to SettingsViewV2? I also need access to currentVersion in the “Report a Problem” screen, so this would help avoid duplication.

giorgikh93 (changes requested):

giorgikh93 (approved):

boris91 (approved):

PR #460, opened 2026-04-28
UI ux fixes

shavtvalishvili:
## Summary

Small UI/UX fixes across the v2 redesign on desktop.

## Changes
- Sort folders alphabetically in folder dropdowns.
- Fix brightness pop on modal close by syncing dim with overlay fade.
- Add scroll fade gradient at the bottom of overflowed move and delete modal item lists.
- Keep multi-select Move and Favorite tooltips inside the viewport.
- Drop redundant protection prompt in the share vault modal. The protected-vaults feature is upcoming (it was present earlier but pulled because of issues), so the check inside `AddDeviceModalContentV2` is currently dead code: every caller operates on the active vault, which is always already authorized. `VaultSelector.handleInvite` is the one path that gates on `isVaultProtected`, and it does so before opening the modal — so once protected vaults relands, the call-site gate is the right place for it.
- Hide the favorite badge inside the move-folder list per design — the dense item rows in this modal intentionally omit the badge.

### Screenshots/Recordings

https://github.com/user-attachments/assets/fc5115f8-80df-4f85-b43c-8c9b066453d3

https://github.com/user-attachments/assets/be5c7270-ba2c-4e4f-b495-620e07842531

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "@tetherto/pearpass-lib-data-export": "git+https://github.com/tetherto/pearpass-lib-data-export.git",
         "@tetherto/pearpass-lib-data-import": "git+https://github.com/tetherto/pearpass-lib-data-import.git",
         "@tetherto/pearpass-lib-native-messaging-bridge": "git+https://github.com/tetherto/pearpass-lib-native-messaging-bridge.git",
-        "@tetherto/pearpass-lib-ui-kit": "git+https://github.com/tetherto/pearpass-lib-ui-react-native-components.git#design-system-v2",
+        "@tetherto/pearpass-lib-ui-kit": "github:tetherto/pearpass-lib-ui-react-native-components#design-system-v2",
         "@tetherto/pearpass-lib-ui-theme-provider": "git+https://github.com/tetherto/pearpass-lib-ui-theme-provider.git",
         "@tetherto/pearpass-lib-vault": "git+https://github.com/tetherto/pearpass-lib-vault.git",
         "@tetherto/pearpass-lib-vault-core": "git+https://github.com/tetherto/pearpass-lib-vault-core.git",
@@ -204,6 +204,7 @@
       "version": "7.27.3",
       "resolved": "https://registry.npmjs.org/@babel/helper-annotate-as-pure/-/helper-annotate-as-pure-7.27.3.tgz",
       "integrity": "sha512-fXSwMQqitTGeHLBC08Eq5yXz2m37E4pJX1qAU1+2cNedz/ifv/bVXft90VeSav5nFO61EcNgwr0aJxbyPaWBPg==",
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "@babel/types": "^7.27.3"
@@ -232,6 +233,7 @@
       "version": "7.28.6",
       "resolved": "https://registry.npmjs.org/@babel/helper-create-class-features-plugin/-/helper-create-class-features-plugin-7.28.6.tgz",
       "integrity": "sha512-dTOdvsjnG3xNT9Y0AUg1wAl38y+4Rl4sf9caSQZOXdNqVn+H+HbbJ4IyyHaIqNR6SW9oJpA/RuRjsjCw2IdIow==",
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "@babel/helper-annotate-as-pure": "^7.27.3",
@@ -253,6 +255,7 @@
       "version": "7.28.5",
       "resolved": "https://registry.npmjs.org/@babel/helper-create-regexp-features-plugin/-/helper-create-regexp-features-plugin-7.28.5.tgz",
       "integrity": "sha512-N1EhvLtHzOvj7QQOUCCS3NrPJP8c5W6ZXCHDn7Yialuy1iu4r5EmIYkXlKNqT99Ciw+W0mDqWoR6HWMZlFP3hw==",
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "@babel/helper-annotate-as-pure": "^7.27.3",
@@ -296,6 +299,7 @@
       "version": "7.28.5",
       "resolved": "https://registry.npmjs.org/@babel/helper-member-expression-to-functions/-/helper-member-expression-to-functions-7.28.5.tgz",
       "integrity": "sha512-cwM7SBRZcPCLgl8a7cY0soT1SptSzAlMH39vwiRpOQkJlh53r5hdHwLSCZpQdVLT39sZt+CRpNwYG4Y2v77atg==",
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "@babel/traverse": "^7.28.5",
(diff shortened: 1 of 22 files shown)


boris91 (dismissed):

torkinos:
@nickgr2 please review

nickgr2 (commented):
A few questions and minor notes — none clearly blocking, but the protection-check removal in particular deserves confirmation before I approve.

### Questions

**1. `AddDeviceModalContentV2` protection check — is the call-site claim accurate?**

The description says protection is "now handled at the call site," but of the four callers I can see, only `VaultSelector.handleInvite` actually gates on `isVaultProtected` (via `switchVault`). The others (`Sidebar.handleAddDevice`, `pages/SettingsViewV2/.../YourVaultsContent.openAddDeviceModal`, `pages/SettingsView/.../SettingsDevicesSection.handleAddDevice`) just call `setModal(<AddDeviceModalContentV2 />)` directly.

I think this is probably fine in practice — the modal operates on the active vault, and the active vault is by definition already authorized — i.e. the check was redundant rather than relocated. If that's the reasoning, could the PR description / a comment make that explicit? "Redundant — active vault is always authorized" reads differently from "moved to call site."

**2. `MoveFolderModalContentV2`: `isFavorite={!!record.isFavorite}` → `isFavorite={false}` (and the `isFavorite?` field dropped from `MoveFolderRecord`).**

This isn't called out in the description but it's a visible UX change — favorited records no longer render the favorite badge inside the move-folder list. Intentional? If the badge was visually noisy in this dense modal that's reasonable, but worth either keeping the badge or noting the rationale.

(Side note: the `useScrollOverflow` comment specifically mentions absolutely-positioned favorite badges as the reason it can't trust `scrollHeight`. After this change there are no badges in this list anyway, but the hook still has to handle them in the delete modal where favorites do render — so the logic is still needed regardless.)

### Minor

- `src/strict.css` — the `*:has(> div > [role='dialog']) > div:first-child` selector targeting the kit's backdrop is brittle: if `Dialog.tsx` ever reorders its children or wraps the backdrop differently, this silently stops working and the double-dim returns. Worth raising a `hideBackdrop` / `disableBackdrop` prop with the ui-kit team so you can drop the `!important` global override. The comment in the file is good — extending it with "raise upstream" as a TODO would help the next reader.
- `src/components/Overlay/styles.js` — the new `rgba(8, 10, 5, 0.72)` is a hardcoded literal (as was the previous value, so this isn't a regression). If the kit exposes a token for "modal overlay scrim" use it; otherwise a const at the top of the file with the design intent would be friendlier than the inline literal.
- `src/utils/withAlpha.ts` — the function throws if the input isn't `#RRGGBB`. That's clearly intentional ("Check that the theme token is defined for the current theme"), and the only current caller passes `colors.colorSurfacePrimary`, which resolves to a 6-digit hex in both the dark and light themes I checked. Just flagging that any future theme token change to `rgba(...)` / 3-digit hex / named color would crash the modal at render — `colorSurfacePrimary` is on a hot path. A console.error + safe fallback (`hex` returned unchanged) might be friendlier, but throwing-is-explicit is also a defensible choice.
- `src/hooks/useScrollOverflow.ts` — non-trivial hook (last-child geometry, ResizeObserver + MutationObserver, deps array forwarded into `useLayoutEffect`) and no tests added. Given it has a tricky invariant ("absolutely-positioned descendants inflate `scrollHeight`") that's easy to regress, a small unit test pinning the last-child measurement strategy would be worth its weight.
- `src/hooks/useAnimatedVisibility.js` — nice fix; the prior version leaked the close `setTimeout` if `isOpen` flipped back to `true` mid-close. Existing tests still pass because they don't pass `nodeRef`/`propertyName`, but consider adding a transitionend-path test so the new branch isn't regression-bait.
- `src/context/ModalContext.js` — switching from `pop()` to id-based removal is the right fix for racy close→open sequences. The new `STACK_CLEANUP_BUFFER = 100` is plumbed through tests, but it's not obvious from the constant alone *why* 100ms — a one-liner ("safety pad past `transitionend`; matches `useAnimatedVisibility` `SAFETY_BUFFER`") would tie the two together.

Overall the changes are tidy and the modal-close fade desync fix is a real improvement. Happy to re-review once the call-site / favorites questions are answered.

shavtvalishvili:
Thanks @nickgr2 — addressed in `54f5b4a` and `e205e8c`. Brief responses:

**Q1 — protection check.** Updated PR description. Protected vaults is an upcoming feature that was previously live but pulled due to issues, so the in-modal check is dead code today. `VaultSelector.handleInvite` is the live gate, and the call site is the right place for it once protected vaults relands.

**Q2 — `isFavorite={false}`.** Intentional, per design — the move-folder list omits the favorite badge. Added an inline comment.

**M1 — `strict.css` selector.** Extended the comment with a TODO to raise `hideBackdrop` upstream so we can drop the `!important` override.

**M2 — hardcoded scrim color.** Went a step further: added `colorScrim` to the ui-kit ([#103](https://github.com/tetherto/pearpass-lib-ui-react-native-components/pull/103), merged), which also migrates the kit's own `Dialog` backdrop. The desktop `Overlay` now reads `theme.colors.colorScrim` via `useTheme()` (`e205e8c`).

**M4 — `useScrollOverflow` tests.** Added 8 tests covering the last-child-vs-scrollHeight invariant, sub-pixel tolerance, ResizeObserver/MutationObserver re-measure, and unmount cleanup.

**M5 — `useAnimatedVisibility` transitionend test.** Added 6 tests for the new branch: matching event, wrong propertyName, descendant bubbling, fallback timer, no double-fire, listener cleanup.

**M6 — `STACK_CLEANUP_BUFFER` comment.** Added a one-liner tying it to `useAnimatedVisibility.SAFETY_BUFFER`.

Ready for re-review.

nickgr2 (approved):
LGTM — all prior questions/notes addressed and the merged-in vault-switch work is clean.

### Verified
- `tertiaryAccent` is a valid `ButtonVariant` in the ui-kit at the lockfile's bumped SHA (`636989931c…`), and `colorScrim` resolves via the dark theme tokens. Light theme is still entirely stubbed (`colorScrim: ''` and many others), but that's a pre-existing kit issue — the Overlay change is parity-preserving since the previous hardcode matched the dark scrim exactly.
- `withRoundedBorders` is a real `ListItem` prop. Switching from `withRoundedBottomBorders={false}` to `withRoundedBorders={false}` in `YourVaultsContent` flattens both top and bottom for "other vault" rows — reads as intentional now that the rows are selectable (the active vault row above provides the rounded cap).
- `useVaultSwitch` extraction matches the inline original behaviorally; loading state is cleaned up correctly across the password-modal hop. The added try/catch logs and rethrows, so callers can still surface failures.
- The new `useScrollOverflow` tests pin the exact invariant the hook exists to protect (absolutely-positioned descendants must not trip overflow), plus ResizeObserver/MutationObserver re-measure paths and unmount cleanup. The new `useAnimatedVisibility` tests cover the matching-property check, descendant bubbling, fallback timer, and listener cleanup.

### Minor (non-blocking)
- `useVaultSwitch` always calls `setIsLoading(true)` up front, including when `vault.id === activeVault?.id` (the no-op path). It still ends in `setIsLoading(false)` via the finally, so it's correct, just briefly toggles a global spinner for what is effectively a synchronous early-return. Could short-circuit before the loading flip.
- `VaultSelector` and `YourVaultsContent` both call `useVaultSwitch()` and so each instantiates its own `useVault()` (the hook destructures `data: activeVault`, `isVaultProtected`, `refetch`). If `useVault` is already cached/shared by a context or query layer this is free; if it triggers extra subscriptions per consumer, worth checking.

Approving.

torkinos (approved):

boris91 (approved):

PR #461, opened 2026-04-28
[v2] the "current vault" experience

giorgikh93:
### Requirements

Header copy update: Change "Your Vaults" section text to: "Manage your vaults. Select the vault you want to apply changes to."
UI Constraint
Display action icons (Sharing and 3-dots menu) only for the currently active vault in both the navbar dropdown (left side) and in the settings → "your vaults" list
Hide all action icons for inactive vaults
Interaction: Ensure all inactive vault rows are fully clickable.
Navigation: Clicking an inactive vault must switch the active context to that vault
Experience: The vault switch must be seamless,  the user must remain on the current page/view without a reload or redirect, but the new "current vault" should be reflected to the selected one
Validation: Confirm the new active vault now displays the Sharing and 3-dots menu, while the previous vault hides them.
### Changes

- Removed rightElement from the inactive vault options in dropdown
- Made ListItem as selectable in "Your Vaults" screen to navigate to the current selected vault
- Update "Your Vaults" text
### Testing Notes

- Tested Locally 
### Screenshots/Recordings

https://github.com/user-attachments/assets/688e63c2-f4b4-4773-abe7-f166216e27e5

### dependencies

---
  - https://app.asana.com/0/0/1214297232978653

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -204,6 +204,7 @@
       "version": "7.27.3",
       "resolved": "https://registry.npmjs.org/@babel/helper-annotate-as-pure/-/helper-annotate-as-pure-7.27.3.tgz",
       "integrity": "sha512-fXSwMQqitTGeHLBC08Eq5yXz2m37E4pJX1qAU1+2cNedz/ifv/bVXft90VeSav5nFO61EcNgwr0aJxbyPaWBPg==",
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "@babel/types": "^7.27.3"
@@ -232,6 +233,7 @@
       "version": "7.28.6",
       "resolved": "https://registry.npmjs.org/@babel/helper-create-class-features-plugin/-/helper-create-class-features-plugin-7.28.6.tgz",
       "integrity": "sha512-dTOdvsjnG3xNT9Y0AUg1wAl38y+4Rl4sf9caSQZOXdNqVn+H+HbbJ4IyyHaIqNR6SW9oJpA/RuRjsjCw2IdIow==",
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "@babel/helper-annotate-as-pure": "^7.27.3",
@@ -253,6 +255,7 @@
       "version": "7.28.5",
       "resolved": "https://registry.npmjs.org/@babel/helper-create-regexp-features-plugin/-/helper-create-regexp-features-plugin-7.28.5.tgz",
       "integrity": "sha512-N1EhvLtHzOvj7QQOUCCS3NrPJP8c5W6ZXCHDn7Yialuy1iu4r5EmIYkXlKNqT99Ciw+W0mDqWoR6HWMZlFP3hw==",
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "@babel/helper-annotate-as-pure": "^7.27.3",
@@ -296,6 +299,7 @@
       "version": "7.28.5",
       "resolved": "https://registry.npmjs.org/@babel/helper-member-expression-to-functions/-/helper-member-expression-to-functions-7.28.5.tgz",
       "integrity": "sha512-cwM7SBRZcPCLgl8a7cY0soT1SptSzAlMH39vwiRpOQkJlh53r5hdHwLSCZpQdVLT39sZt+CRpNwYG4Y2v77atg==",
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "@babel/traverse": "^7.28.5",
@@ -339,6 +343,7 @@
       "version": "7.27.1",
       "resolved": "https://registry.npmjs.org/@babel/helper-optimise-call-expression/-/helper-optimise-call-expression-7.27.1.tgz",
       "integrity": "sha512-URMGH08NzYFhubNSGJrpUEphGKQwMQYBySzat5cAByY1/YgIRkULnIy3tAMeszlL/so2HbeilYloUmSpd7GdVw==",
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "@babel/types": "^7.27.1"
@@ -351,6 +356,7 @@
       "version": "7.28.6",
       "resolved": "https://registry.npmjs.org/@babel/helper-plugin-utils/-/helper-plugin-utils-7.28.6.tgz",
       "integrity": "sha512-S9gzZ/bz83GRysI7gAD4wPT/AI3uCnY+9xn+Mx/KPs2JwHJIz1W8PZkg2cqyt3RNOBM8ejcXhV6y8Og7ly/Dug==",
+      "devOptional": true,
       "license": "MIT",
       "engines": {
         "node": ">=6.9.0"
@@ -378,6 +384,7 @@
       "version": "7.28.6",
       "resolved": "https://registry.npmjs.org/@babel/helper-replace-supers/-/helper-replace-supers-7.28.6.tgz",
       "integrity": "sha512-mq8e+laIk94/yFec3DxSjCRD2Z0TAjhVbEJY3UQrlwVo15Lmt7C2wAUbK4bjnTs4APkwsYLTahXRraQXhb1WCg==",
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "@babel/helper-member-expression-to-functions": "^7.28.5",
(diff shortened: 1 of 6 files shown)


giorgikh93:
@nickgr2 please review

boris91 (dismissed):

nickgr2 (changes requested):
A few blocking issues in `src/pages/SettingsViewV2/content/YourVaultsContent/index.tsx`. Re-review once these are addressed.

### 1. Conditional hook (Rules of Hooks violation)

```tsx
if (!vault) {
  return null            // line 111-113 — early return
}

const switchToVault = useCallback(async (v: Vault) => {
  ...
}, [vault?.id])          // line 115 — hook called AFTER the early return
```

This violates the rules of hooks: when `vault` is initially undefined the component renders with N hooks, and the next render (once `vault` resolves) calls N+1 hooks. React will throw "Rendered more hooks than during the previous render". Move the `useCallback` above the `if (!vault)` guard along with the other hooks.

### 2. Invalid `ListItem` prop — `withRoundedBorders` does not exist

```tsx
<ListItem
  ...
  withRoundedBorders={false}
```

The kit's `ListItem` exposes [`withRoundedBottomBorders`](node_modules/@tetherto/pearpass-lib-ui-kit/src/components/ListItem/ListItem.tsx#L34) — not `withRoundedBorders`. The previous code used `withRoundedBottomBorders={false}`. As-written the prop is ignored, `withRoundedBottomBorders` falls back to its `true` default, and every "Other Vaults" row will render with rounded bottoms instead of the intended flat-bottom stack. Restore the original prop name.

### 3. Missing protected-vault password prompt

`switchToVault` calls `refetchVault(v.id)` directly. The existing `VaultSelector.switchVault` (`src/containers/Sidebar/VaultSelector/VaultSelector.tsx`) intentionally checks `isVaultProtected(vault.id)` first and routes through `<VaultPasswordFormModalContent>` so the user can supply the password — without that, `refetchVault` for a protected vault either errors or leaves redux in an inconsistent state. The Settings flow needs the same gating, otherwise users can't switch to any protected vault from this surface.

### 4. `useCallback` deps incomplete

`[vault?.id]` is missing `refetchVault` and `setIsLoading`. `react-hooks/exhaustive-deps` will warn. (`vault` itself is also referenced — `vault?.id` already covers identity, but include the others.)

### 5. minor: copy

`'Manage your vaults. Select the vault you want to apply changes to.'` — selecting a row here actually switches the *active* vault globally, not just the surface where edits land. Worth aligning the copy with the behavior to avoid surprising the user.

### 6. minor: formatting noise unrelated to the feature

- Indentation regression on `count: vault?.devices?.length ?? 0` inside the `DevicesMeta` object literal (lines 105-108).
- Two blank lines after the `useTheme()` block (45-46) and after the new `useCallback` (129-130).

A Prettier pass should clean these up.

### Note on `VaultSelector.tsx`

`rightElement={isActive ? rightElement : undefined}` is correct for the spec. No issues there.

alvaropaco (dismissed):

giorgikh93:
@nickgr2 please review again! I addressed some of your points and provided reasoning for those items I didn't address.

1. Addressed this point
2. `withRoundedBorders` is actually exists and you can check node_modules/@tetherto/pearpass-lib-ui-kit/dist/components/ListItem/ListItem.d.ts line 30  ` withRoundedBorders?: boolean`
3. Protected vault flow is out of scope, we can skip this.
4. Added setIsLoading in dependency but didn't add refetchVault because refetchVault from useVault() isn’t referentially stable, it’s often a new function each render. If we list it in the effect deps, the effect re-runs every render, calls refetch again, state updates, re-render, new refetch - loop. We only need to re-run when something meaningful changes (e.g. recordsData?.length), so we don’t add refetchVault to the deps unless the library guarantees a stable refetch (e.g. useCallback in the provider).
5. we can skip this as it's product requirement.
6. addressed

nickgr2:
> `withRoundedBorders` is actually exists and you can check node_modules/@tetherto/pearpass-lib-ui-kit/dist/components/ListItem/ListItem.d.ts line 30

It doesn't. `pearpass-lib-ui-kit` doesn't ship a `dist/` directory locally (the kit is consumed from source via `git+...#design-system-v2`), and the only rounded-borders prop in `node_modules/@tetherto/pearpass-lib-ui-kit/src/components/ListItem/ListItem.tsx` is `withRoundedBottomBorders` (line 34). Line 30 in that file is `titleStyle?: TextProps['style']`, not `withRoundedBorders`. Grep across the entire kit package returns zero matches for `withRoundedBorders`. As-written, `withRoundedBorders={false}` is silently dropped, the default `withRoundedBottomBorders=true` kicks in, and "Other Vaults" rows render with rounded bottoms instead of the flat-bottom stack the previous code had. Please restore `withRoundedBottomBorders={false}`.

> refetchVault from useVault() isn't referentially stable

Confirmed in `node_modules/@tetherto/pearpass-lib-vault/src/hooks/useVault.js` — `refetch` is a fresh closure on every render. The `useCallback` deps as-written are fine; an upstream `useCallback` on `refetch` would be the cleaner fix but is out of scope here.

nickgr2 (changes requested):
Thanks for the fixes — hook order is now correct and the formatting noise is cleaned up. Two of the original blockers still stand at `b81c38aa`.

### 1. `withRoundedBorders` is still invalid (was #2)

```tsx
withRoundedBorders={false}
```

I re-verified at this SHA against the locked kit (`@tetherto/pearpass-lib-ui-kit` resolved from the `design-system-v2` branch, currently `0.0.40` locally). Only `withRoundedBottomBorders?: boolean` is defined in `node_modules/@tetherto/pearpass-lib-ui-kit/src/components/ListItem/ListItem.tsx` (line 34, default `true`, consumed at line 123). There is no `withRoundedBorders` prop — TS extra-prop checking accepts it via JSX spread, but at runtime it is dropped, `withRoundedBottomBorders` keeps its `true` default, and every "Other Vaults" row renders with rounded bottoms instead of the intended flat-bottom stack inside `MultiSlotInput`. Rename to `withRoundedBottomBorders={false}`.

### 2. Switching to a protected vault still bypasses the password prompt (was #3)

```tsx
const switchToVault = useCallback(
  async (v: Vault) => {
    if (v.id === vault?.id) return
    try {
      setIsLoading(true)
      await refetchVault(v.id)
    ...
  },
  [vault?.id, setIsLoading]
)
```

The sibling flow at `src/containers/Sidebar/VaultSelector/VaultSelector.tsx` (lines 54, 92, 96) calls `isVaultProtected(vault.id)` first and routes through `<VaultPasswordFormModalContent>` before unlocking. `YourVaultsContent` calls `refetchVault(v.id)` unconditionally, so clicking a protected vault row from Settings either errors or leaves redux inconsistent — the user has no way to supply the password from this surface. Reuse the same gating (extract a shared helper or call `VaultSelector`'s `switchVault` path) so both entry points behave the same.

### Minor (was #5, optional)

`'Manage your vaults. Select the vault you want to apply changes to.'` — selecting a row here switches the *active* vault globally, not just where edits land. Worth aligning the copy with the behavior, but won't block.

giorgikh93:
@nickgr2 please review, I've addressed your comments except withRoundedBorders. Please fetch new dependencies, I assume you have a stale dependencies now.

nickgr2 (approved):
Looks good — issue #3 (password-modal bypass in `YourVaultsContent`'s `switchToVault`) is properly addressed by extracting `useVaultSwitch` and routing both call sites through it. The hook correctly checks `isVaultProtected` and opens `VaultPasswordFormModalContent` before refetching, so clicking an inactive protected vault from the Settings list now goes through the same flow as the dropdown.

The unit tests on `useVaultSwitch` are thorough — covering same-vault, unprotected, protected (modal opens, no premature refetch), modal `onSubmit` (refetch with password + close + onSuccess), and the rejected-refetch logging path.

Re #2 (`withRoundedBorders`): you were right, I was wrong. I re-read the kit on the `design-system-v2` branch via the GitHub API and the prop is exported (`ListItem.tsx` line 36/67/129). My previous reviews trusted my local `node_modules/@tetherto/pearpass-lib-ui-kit/` which was out of sync with the remote branch. Sorry for the back-and-forth on that one.

One minor non-blocker for future cleanup: the four call sites in `VaultSelector` use `void switchVault(...)` while the hook catches → logs → rethrows. With `void`, the rethrow has no consumer and produces an unhandled promise rejection. Either drop the rethrow (log only) or catch at the call sites — but this is a code-smell, not a regression vs. the prior inline implementation.

Approving.

alvaropaco (approved):

boris91 (approved):

PR #463, opened 2026-04-29
Use v2 modal for deleting record

ndolidzee:
### Changes

Use v2 delete records modal when deleting record

### Screenshots/Recordings

https://github.com/user-attachments/assets/a0f9cccc-6dcd-4791-8963-05479acef98c

https://github.com/user-attachments/assets/13086a3d-d893-4322-94bd-f640e131f2cc

### DIFF
--- a/src/hooks/useRecordActionItems.js
+++ b/src/hooks/useRecordActionItems.js
@@ -4,6 +4,7 @@ import { html } from 'htm/react'
 
 import { useCreateOrEditRecord } from './useCreateOrEditRecord'
 import { ConfirmationModalContent } from '../containers/Modal/ConfirmationModalContent'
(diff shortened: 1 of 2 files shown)


ndolidzee:
@nickgr2 please review

nickgr2 (approved):
minor: `handleMoveClick` in this file still uses the `html`` template tag while the new `handleDelete` uses raw JSX. Same file, mixed styles. Not a blocker — worth tidying when convenient.

Otherwise the V2 wiring matches the established pattern from `RecordRowContextMenuV2.tsx:152` and `MainViewV2.tsx:136`, and the test split between the v2 default (`isV2 → true`) and the explicit v1 override for the confirmation test reads cleanly.

boris91 (approved):

torkinos (approved):

PR #464, opened 2026-04-29
handle smooth navigation

giorgikh93:
### Changes

- Smooth vault navigation from sidebar - the idea is to not close the sidebar after navigation
- remove unnecessary vault navigation calls while clicking rename or open pairing flow
### Screenshots/Recordings

https://github.com/user-attachments/assets/01cbe61c-7739-4bba-90fe-1d5bfbf5edab

### DIFF
--- a/src/containers/AppHeaderContainer/AppHeaderContainer.test.js
+++ b/src/containers/AppHeaderContainer/AppHeaderContainer.test.js
@@ -28,6 +28,12 @@ jest.mock('../../context/ModalContext', () => ({
   })
 }))
 
+jest.mock('../../hooks/useTranslation', () => ({
+  useTranslation: () => ({
+    t: (key) => key
+  })
+}))
+
 jest.mock('../../hooks/useRecordMenuItems', () => ({
   useRecordMenuItems: jest.fn()
 }))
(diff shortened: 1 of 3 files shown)


giorgikh93:
@nickgr2 please review

nickgr2 (approved):
Reviewed the full diff at `7e3e821` (queued at `39507b9` plus the test-mock fix in `7e3e821`).

The cleanup is correct: `handleInvite`, `handleRename`, `handleSetPassword`, and `handleDelete` are reachable only via the row's `rightElement`, which renders only when `isActive` (`VaultSelector.tsx:291`). Since `useVaultSwitch`'s `switchVault` early-returns for the same vault, the removed wrappers around `openInviteModal` and the rename modal were no-ops on the active vault — dropping them is a clarity/perf win without behavior change. `createInvite` reads `getState().vault.data.id`, which still resolves to the intended vault because the action is gated on active.

Removing the `onClose` from `<VaultSelector />` matches the PR description — the chevron toggle (`SidebarV2.tsx:225`) and the explicit close button (`SidebarV2.tsx:204`) still let the user collapse the selector, so no UX regression.

The `useTranslation` mock added to `AppHeaderContainer.test.js` is required because the component calls `t(...)` for the category labels (`AppHeaderContainer.tsx:42`,`66-75`); the prior test was relying on Jest's auto-mocking and would have surfaced as a `t is not a function` error in CI.

minor: `handleSetPassword` still wraps the `setModal` call in `switchVault(vault, …)`. Given the same active-vault gating, that path also reduces to `setModal(...)` after the early return — worth flattening for consistency with the other handlers, but not a blocker.

boris91 (approved):

torkinos (approved):

PR #465, opened 2026-04-29
Adjust hidden messages in the new record items

denislupookov:
### Changes

[v2] [desktop] Adjust hidden messages in the new record items

custom field (BE) would be refelcted as "hidden message" in the app
"other" type would only have title and hidden message wihtout comments

---
  - https://app.asana.com/0/0/1214386530275803

### DIFF
--- a/src/containers/Modal/CreateOrEditCategoryWrapper/CreateOrEditCustomModalContentV2/CreateOrEditCustomModalContentV2.tsx
+++ b/src/containers/Modal/CreateOrEditCategoryWrapper/CreateOrEditCustomModalContentV2/CreateOrEditCustomModalContentV2.tsx
@@ -9,6 +9,7 @@ import {
   Form,
   InputField,
   MultiSlotInput,
+  PasswordField,
   Text,
   useTheme
 } from '@tetherto/pearpass-lib-ui-kit'
@@ -104,7 +105,7 @@ export const CreateOrEditCustomModalContentV2 = ({
       title: initialRecord?.data?.title ?? '',
       customFields: initialRecord?.data?.customFields?.length
         ? initialRecord.data.customFields
-        : [{ type: 'note', note: '' }],
+        : [{ type: 'note', name: 'note', note: '' }],
       folder: selectedFolder ?? initialRecord?.folder,
       attachments: initialRecord?.attachments ?? []
     },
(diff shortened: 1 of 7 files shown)


denislupookov:
@nickgr2 please review!

boris91 (approved):

nickgr2 (approved):
The change cleanly migrates custom fields to `PasswordField` ("Hidden Message") across the four create/edit modals and the three detail forms, drops the comments section from `CreateOrEditCustomModalContentV2` per the PR description, and routes everything through the kit + Lingui. Renderer hardening, kit imports, and `testID` coverage all look good.

A few minor consistency notes — none blocking:

**`addCustomField` initial shape differs across modals**
- `CreateOrEditCustomModalContentV2.tsx` (and the existing `CreateOrEditLoginModalContentV2.tsx`) pass `{ type: 'note', name: 'note', note: '' }`.
- `CreateOrEditNoteModalContentV2.tsx` and `CreateOrEditPassPhraseModalContentV2.tsx` pass `{ type: 'note', note: '' }` — no `name`.

The schema validates only `note` and the type defs declare `name` as either optional (`Custom`) or required (`Note`/`PassPhrase`), so this isn't a runtime bug, but the four modals now persist subtly different shapes for what is conceptually the same record. Worth picking one shape and using it everywhere so downstream consumers don't hit surprises.

**`testID` casing inconsistency for the new section**
- Modal slots/inputs use `hiddenmessage` (one token): `createoredit-{type}-input-hiddenmessage-v2-{i}`, `createoredit-{type}-hiddenmessage-slot-v2`.
- Detail forms use `hidden-messages` (two kebab tokens): `hidden-messages-multi-slot-input`, `hidden-messages-multi-slot-input-slot-{i}`.

Pre-existing detail form testIDs (e.g. `comments-multi-slot-input`) were kebab-cased, so the detail-form choice is consistent with the file's existing style — but it makes E2E selectors slightly harder to predict. Worth flagging in case the QA selectors converge later.

**`PassPhrase` modal seeds `customFields` to `[{ type: 'note', note: '' }]` even when none existed.**
Previously the form defaulted to `[]`, so a brand-new recovery-phrase record had no hidden-message rows. After this change, every new record always shows one empty `PasswordField`. The new `.filter(f => f.note?.trim().length)` on submit means empty rows are stripped before saving, so there's no data leak — just a UX choice to expose the input by default. Same pattern as `Note`/`Custom` modals, so probably intentional.

minor: `onSubmit` casts `formValues.customFields` via `as Array<{ type: string; note?: string }>` to call `.filter(...)`. Given `formValues` is `Record<string, unknown>` from the form lib, the assertion is reasonable; just be aware the cast diverges from the declared `{ type: string; name: string }[]` (Note/PassPhrase) and `{ type: string; name?: string; note?: string }[]` (Custom) prop types — another argument for converging on one shared `CustomField` type.

torkinos (approved):

PR #466, opened 2026-04-29
[v2] [desktop] Locked screen

giorgikh93:
### Screenshots/Recordings

Screenshot 2026-04-29 at 18 18 22

### dependencies

---
  - https://app.asana.com/0/0/1214386530275810

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6218,7 +6218,7 @@
     "node_modules/@tetherto/pearpass-lib-ui-kit": {
       "name": "pearpass-lib-ui-react-native-components",
       "version": "0.0.40",
-      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#636989931c65a63d4dd429911facbf38d60d1fc2",
+      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#2499e063482020abd394eadb9b878b8126f22d99",
       "license": "Apache-2.0",
       "dependencies": {
         "react-strict-dom": "^0.0.55"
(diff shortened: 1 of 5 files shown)


giorgikh93:
@nickgr2 please review

nickgr2 (commented):
CI is green and the change is small and well-scoped. One styling nit, otherwise this looks good.

**`minor:` Hardcoded `borderRadius: '10px'`**

[`LockedScreenV2.styles.ts:76`](https://github.com/tetherto/pearpass-app-desktop/pull/466/files#diff-LockedScreenV2.styles.ts) — the pill uses

```ts
borderRadius: '10px',
```

Every other v2 surface I checked goes through tokens (`SettingsViewV2/.../*.styles.ts` all use `${rawTokens.radius8}px`), and `AGENTS.md` calls out radius as a token-only value. If 10px is a deliberate spec value not in the token set (`radius8 | radius16 | radius20 | radius26`), worth confirming with design; otherwise switch to `${rawTokens.radius8}px` for consistency.

**Other things I checked (all good)**

- `WatchLater` icon — verified it lands in the kit bump (`2499e06...src/icons/components/WatchLater.tsx`); the `0.0.40 → 0.0.40` lockfile-only bump is intentional.
- `V2` suffix is correct (sibling `LockedScreen/` exists).
- All user-facing strings go through `t(...)`, all interactive/queryable nodes have `testID`/`data-testid`.
- `useEffect` deps array now includes `refreshMasterPasswordStatus` (v1 had `[]`) — that's the correct fix.
- `OnboardingShell background="solid"` matches the existing prop contract.
- Tests cover the four meaningful states (loading placeholder → countdown, ceil-on-ms math, finish-then-unlocked navigates, finish-then-still-locked stays). `Math.ceil(0/1000) === 0` placeholder branch is exercised — nice.

giorgikh93:
@nickgr2 addressed your points, please review again

nickgr2 (approved):
Token nit addressed — `borderRadius: \`${rawTokens.spacing10}px\`` keeps the 10px design value and routes it through a token. There's no `radius10` in the kit (radius8/16/20/26) and the `spacing*`-for-radius pattern has precedent (`ImportItemsContent`, `GeneratePasswordModalContentV2`, `PassPhraseV2` all use `spacing8` the same way), so this reads fine.

LGTM.

boris91 (approved):

hakaari (approved):

PR #469, opened 2026-04-30
Add strict-confinement snap with native messaging for host and snap browsers

shavtvalishvili:
- New `snap/snapcraft.yaml`: strict-confinement snap, `core22` base, two apps (`pearpass` GUI + `native-host` slot the browser execs as `/snap/bin/pearpass.native-host`), `gnome` extension, `personal-files` plug (`browser-native-messaging`). Snap version is sourced from `package.json` via `craftctl set version=...` (using `adopt-info: pearpass` without `parse-info`), so bumping `package.json` is the only action needed at release time.
- New `scripts/build-snap.sh`: wraps `snapcraft pack`. Two staging modes: `--unpacked <dir>` (preferred, uses electron-builder's `linux-*-unpacked` directly) and `--local <AppImage>` (fallback). Globs the produced snap to absorb `--output` flag inconsistency across snapcraft versions.
- `electron/flatpak-paths.cjs`: add `isSnapRuntime()` and `getSnapRealHome()` (mirrors `isFlatpakRuntime`/`getHostHome`); honors `SNAP_REAL_HOME` for the host-home redirect.
- `src/utils/nativeMessagingSetup.js`: snap branch in `setupNativeMessaging` writes the manifest's `path` as `/snap/bin/pearpass.native-host` directly (no wrapper, since other snaps' browsers can't reach our user-data dir). New `Firefox (Snap)` browser entry covers `~/snap/firefox/common/.mozilla/native-messaging-hosts/`. Existing `Chromium (Snap)` and `Brave (Snap)` paths get the same treatment.
- `personal-files` is **directory-scoped**: file scope can't `mkdir` the parent `NativeMessagingHosts/`, which neither Firefox nor Chromium-family browsers create on their own (verified empirically against host + snap Firefox 150 and against `NativeManifests.sys.mjs` source: only `IOUtils.readJSON`, no writes).
- Removes the prior electron-builder `snap` block from `package.json` (it was `confinement: classic`; superseded by `snap/snapcraft.yaml`).
- Smoke-tested on Ubuntu 26.04 aarch64 with snap-Firefox and Mozilla's host Firefox 150: pairing and native messaging round-trip both work.
- Until the snap store grants auto-connect via signed snap-declaration, end users must run `sudo snap connect pearpass:browser-native-messaging` once after install.

## Release-time process

1. Bump `package.json`. The snap version is picked up automatically.
2. Prepend a new `<release version="X.Y.Z" date="YYYY-MM-DD">` block to `flatpak/com.pears.pass.metainfo.xml` with a short user-facing changelog. AppStream renders it as the changelog in GNOME Software, KDE Discover, snapcraft.io, and flathub.


shavtvalishvili:
Companion CI change: tetherto/pearpass-apps-releases#113 - updates `build-desktop-app.yaml`'s `build-snap` job to use `./scripts/build-snap.sh --unpacked` instead of the now-removed `build.snap` block in `package.json`. The two should merge together (or this PR first, then 113); CI will fail on `dist:snap:*` after this PR lands without #113.

hakaari:
@nickgr2 can you please review?

flatpak/com.pears.pass.metainfo.xml
```diff
@@ -62,6 +62,13 @@
...
 
   <releases>
+    <!--
+      Prepend a new <release> entry (newest first) on every release.
+      AppStream renders this as the user-visible changelog in GNOME Software,
+      KDE Discover, snapcraft.io, and flathub. The snap version itself is
+      sourced from package.json via craftctl in snap/snapcraft.yaml, so the
+      list here is for changelog content only — not a build input.
+    -->
     <release version="1.0.1" date="2024-12-19">
```
hakaari, line 72: should we include 2.0.0 release info here as well?

nickgr2 (approved):
This is a well-structured strict-confinement snap setup that mirrors the flatpak path closely. The new helpers are tested, the manifest path indirection through `/snap/bin/<snap>.native-host` cleanly avoids the user-data wrapper, and dropping the prior `confinement: classic` block in `package.json` is a meaningful security improvement.

A few minor observations / questions, none blocking:

1. **`--disable-gpu` missing from snap launcher** — `snap/snapcraft.yaml` passes `--no-sandbox --disable-gpu-sandbox --disable-dev-shm-usage --enable-features=UseOzonePlatform --ozone-platform=x11`, but the flatpak launcher in `flatpak/com.pears.pass.yaml` also passes `--disable-gpu`. The flatpak side added it because Wayland/GPU detection produced a black window on CI bundles — is the snap reliably avoiding that under XWayland on `core22`+`gnome` extension, or worth carrying the same flag for parity?

2. **`browser-native-messaging` plug is directory-scoped across 8 dirs (including other snaps' data dirs)** — the inline comment correctly notes file-scoped paths break browser mkdir of `NativeMessagingHosts/`, and that this requires manual `sudo snap connect`. Worth being explicit somewhere user-facing (README / release notes?) that the integration won't work until that connection is made — `build-snap.sh`'s final log mentions it, but end users installing from `.snap` artifacts later won't see that output. Snap Store auto-connect of `personal-files` also tends to need a store review.

3. **`extensions: [gnome]` on the `native-host` app** — the host runs as a stdio child with no GUI; if this is needed because the Electron binary still loads GTK at startup, a one-line comment justifying it (mirroring the existing `--no-sandbox` and stage-packages comments) would make it easier to tell whether it can be slimmed later.

4. **`metainfo.xml` changelog drift** — the new comment block ("Prepend a new `<release>` entry on every release") is helpful, but the most recent `<release>` is still `1.0.1` (2024-12-19) while `package.json` is at `1.6.0`. Pre-existing, not caused by this PR — flagging only because the new comment will draw reviewers' eyes here. Could be batched into a follow-up release-notes pass.

5. **`isSnapRuntime()` only checks `SNAP_NAME`** — fine and consistent with `isFlatpakRuntime()`'s minimal detection. Just noting that snapd also exports `SNAP_INSTANCE_NAME` / `SNAP_REVISION`, so if you ever want a stricter check, multiple-var match is an option.

minor: snap-side wrapper writes `${SNAP}` literally inside the heredoc (correct — must expand at runtime, not at `override-build` time); easy to misread. The single-quoted `<<'WRAPPER'` makes this safe, but a one-line "deliberately runtime-expanded" note would help future readers.

hakaari (dismissed):

torkinos (dismissed):

torkinos (approved):

hakaari (approved):

PR #470, opened 2026-04-30
Automation v2

stefannikolic:


boris91 (dismissed):

e2e/components/CreateOrEditPage.js
```diff
@@ -153,19 +185,41 @@ class CreateOrEditPage {
...
-    return this.root.getByTestId('customfields-button-remove')
+    return this.root.getByTestId('createoredit-custom-button-removecustomfield-v2-0')
   }
 
   get dropdownFolderMenu() {
-    return this.root.getByTestId('createoredit-dropdown-folder')
+    return this.root.getByTestId('createoredit-select-folder-v2')
   }
 
+  // get dropdownFolderMenu() {
```
giorgikh93: Please remove unnecessary comments

e2e/components/CreateOrEditPage.js
```diff
@@ -214,12 +277,20 @@ class CreateOrEditPage {
     return this.root.getByTestId('createoredit-button-save')
   }
 
+  // get fileInput() {
```
giorgikh93: if this is not necessary please remove

e2e/components/DetailsPage.js
```diff
@@ -7,24 +7,95 @@ class DetailsPage {
...
 
+  getItemDetailsSlot(slot) {
+    return this.root.getByTestId(`credentials-multi-slot-input-slot-${slot}`).locator('input')
+  }
+
+  getItemDetailsByPlaceholder(placeholder) {
+    return this.root.getByPlaceholder(placeholder)
+  }
+
+  // async verifyItemDetailsByPlaceholderValue(placeholder, value) {
```
giorgikh93: please remove if it's not necessary

e2e/components/DetailsPage.js
```diff
@@ -7,24 +7,95 @@ class DetailsPage {
...
+    return this.root
+    .getByTestId('details-header-v2')
+    .locator('input[placeholder]')
+  }
+
+  async verifyDetailsNoItems() {
+    await expect(this.itemDetailsCounter).toHaveCount(0)
+  }
+
+  // async verifyDetailsSlotNotVisible(slot) {
```
giorgikh93: same

e2e/playwright.config.js
```diff
@@ -27,7 +27,7 @@ export default defineConfig({
     [
       'playwright-qase-reporter',
       {
-        mode: 'testops',
+        mode: process.env.QASE_MODE ?? 'off',
```
giorgikh93: I don't see where you set this env variable, is it necessary?

torkinos:
@nickgr2 please review

nickgr2 (commented):
## Review

This PR migrates Playwright E2E tests from v1 UI selectors to v2, removes the now-obsolete `VaultSelectPage`, and adds an `ELECTRON_RUN_AS_NODE` env scrub. The bulk of the diff is mechanical testID updates and is internally consistent. A few concrete concerns below.

### Dead / commented-out code (please clean up)

There is a significant amount of leftover code throughout the diff that should be deleted rather than kept commented:

- **Page object files** carry alternate locator definitions side-by-side with the active one. Examples: `e2e/components/LoginPage.js` keeps two commented `get title()` and `get passwordInput()` blocks; `e2e/components/DetailsPage.js` has commented duplicates of `getItemDetailsCustomInput`, `verifyCustomNoteText`, `detailsBarFavoriteButton`, `createFolderButton`, `clickFavoriteButton`; `e2e/components/SideMenuPage.js` has three commented copies of `async deleteFolder` and one commented `get sidebarAddButton`; `e2e/components/SettingsPage.js` has a commented `get backSettingsButton`. `MainPage.js` keeps `favoriteIconDisabled`/`favoriteIconEnabled` as well as commented duplicates with the same names. These look like working alternates kept “just in case” — they will rot. Pick the one you want and delete the rest.
- **Test bodies disabled by commenting** rather than `test.skip(...)`/`test.fixme(...)`. Examples: `Add Custom Note` / `Delete Note field` in `creatingLoginItem` and `creatingCreditCardItem`; `View uploaded file in View mode (and cleanup)` across Login/CreditCard/Identity/Note specs; `Dropdown moves to selected item edit screen` and `Item moved to folder (and cleanup)` in CreditCard/WiFi/PassPhrase/Note/CustomField/Identity. If they’re blocked on missing v2 functionality (folder selector in CreditCard/Custom modals etc.), prefer `test.fixme('reason / link to ticket', ...)` so they show up in reports rather than silently disappearing.
- Inside surviving tests, large blocks of replaced code are kept as comments (e.g., `creatingPassPhraseItem.test.js` has nested `// // test(...)` blocks; `editingDeletingIdentityItem.test.js` keeps ~120 lines of commented prior-version assertions inline with the new ones). git history is the right place for this.

### Lost coverage (worth keeping in some form)

- `e2e/components/LoginPage.js#loginToApplication`: the title assertion was commented out (`// await expect(this.title).toHaveText('Enter your Master password')`). `waitForReady` only awaits visibility of the new `h1` locator — if the renderer briefly shows the wrong screen with an `h1`, you’d advance past it. Either restore the assertion against the new copy or rely on a stable testID for the title.
- The “View uploaded file in Edit mode” tests dropped the post-upload preview verification (`verifyUploadedImageIsVisible` / `clickElementItemCloseButton`) before `save()`. Combined with “View uploaded file in View mode (and cleanup)” being commented out across multiple specs, the attachment-preview path and the attachment-delete cleanup are no longer covered for Login/CreditCard/Identity/Note.
- `e2e/specs/01-Login/creatingLoginItem.test.js` “Empty fields not displayed in view mode” replaced four explicit `verifyItemDetailsValueIsNotVisible` assertions with a single `verifyDetailsNoItems()` (count==0 on `details-header-v2 input[placeholder]`). That asserts “no editable inputs in the details header” but no longer pins which fields disappear — if a single empty field starts rendering again, the test still passes. Worth keeping at least one targeted assertion per field type.

### Real bugs

- `e2e/components/MainPage.js` declares `async clickYesButton()` twice in the same class (after the diff: once around the original location, once at the bottom). Both bodies are identical (`getByTestId('delete-records-submit-v2').click()`). The second silently shadows the first — please remove the duplicate.
- `e2e/specs/01-Login/creatingLoginItem.test.js` `beforeAll` and `afterAll` both wrap `sideMenuPage.deleteFolder('Test Folder')` in an empty `try { ... } catch (e) { /* folder may not exist */ }`. Swallowing all errors hides genuine breakages (selector drift, app crashes during teardown, etc.). Prefer a `deleteFolderIfExists` that checks `getSideMenuFolder(name).isVisible()` first and returns cleanly, so unexpected errors still surface.
- `e2e/components/SideMenuPage.js#verifySideBarFavoritesFolder` is asserted with hard-coded strings `'1 items'` / `'0 items'` in many specs. If the app currently emits the literal `"1 items"` as the aria-label, that itself is a UX/i18n bug (should be `"1 item"` via Lingui plurals). If it doesn’t, the assertion would never pass with one element. Either way this string deserves a second look.

### Smaller things

- `e2e/components/DetailsPage.js#markAsFavoriteButton` and `removeFromFavoritesButton` couple to localized text `"Add to Favorites"` / `"Remove from Favorites"` inside the action item testID. Once Lingui starts translating, these tests will fail in non-English runs. If the buttons expose stable aria-labels or distinct testIDs, prefer those.
- `e2e/components/MainPage.js#getElementByPosition` returns `.locator('span').last()`. "Last span inside a record row" is brittle to layout changes; if the row gains a metadata span at the end (timestamp, badge, …), the assertion target shifts silently.
- `e2e/components/MainPage.js#getElementByPosition1` calls `.nth(`${position}`)` with a template-string-coerced value. `nth` expects a number; the diff already fixes this for `getElementByPosition` (passing the bare `position`) but `getElementByPosition1` still uses the template form. Worth normalizing.
- `e2e/playwright.config.js`: switching qase reporter mode to `process.env.QASE_MODE ?? 'off'` is fine, but the giant batch of `// import { qase } from ...` and `// qase.id(N)` comments throughout the spec files suggests the qase reporter is being abandoned for now. If that’s the case, removing the import + calls outright (and dropping the dependency in a follow-up if unused) is cleaner than leaving every test file littered with commented imports.
- `e2e/specs/05-PassPhrase/editingDeletingPassPhraseItem.test.js` indentation is off in the new `beforeAll` (`     await mainPage.clickAddItem('passPhrase')` has a leading double-space).

Happy to re-review once the dead code / coverage gaps are addressed; the underlying v2 selector migration looks sound.

e2e/components/SettingsPage.js
```diff
@@ -5,65 +5,90 @@ class SettingsPage {
...
 
-  // ==== LOCATORS ====
+  // --- Navigation ---
 
-  getSettingsNavigations(navigation_menu_navigation_id) {
-    return this.root.getByTestId(
-      `settings-nav-${navigation_menu_navigation_id}`
-    )
-  }
+  // getSettingsNavigations(navigation_menu_navigation_id) {
```
giorgikh93: is this neceesary to keep this comment?

giorgikh93 (approved):

torkinos (approved):

PR #472, opened 2026-05-01
Create items bugs

giorgikh93:
### Requirements

[v2][MacOS/Windows/Ubuntu][Creating Item] The "Add Item" button is enabled when there are empty fields
[v2][MacOS/Windows/Ubuntu][Item Edit/Create mode] Once an item is placed into a folder, it cannot be made folderless again in edit mode
### Screenshots/Recordings

https://github.com/user-attachments/assets/bd08e3cc-3bd5-480a-840b-6ce8ce1d0870

https://github.com/user-attachments/assets/015f1d29-088e-44ef-a960-34dbd0c3bfc5

### DIFF
--- a/src/components/FolderDropdown/FolderDropdownV2.tsx
+++ b/src/components/FolderDropdown/FolderDropdownV2.tsx
@@ -1,13 +1,15 @@
-import React, { useMemo } from 'react'
+import { useMemo } from 'react'
 
 import {
+  Button,
   ContextMenu,
   MultiSlotInput,
   NavbarListItem,
   SelectField,
+  rawTokens,
   useTheme
 } from '@tetherto/pearpass-lib-ui-kit'
-import { CreateNewFolder, Folder, KeyboardArrowBottom } from '@tetherto/pearpass-lib-ui-kit/icons'
+import { Close, CreateNewFolder, Folder, KeyboardArrowBottom } from '@tetherto/pearpass-lib-ui-kit/icons'
 import { useFolders } from '@tetherto/pearpass-lib-vault'
 
 import { CreateFolderModalContentV2 } from '../../containers/Modal/CreateFolderModalContentV2/CreateFolderModalContentV2'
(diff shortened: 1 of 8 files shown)


giorgikh93:
@nickgr2 please review

torkinos (approved):

nickgr2 (approved):
Both fixes look correct and tightly scoped.

**Folder clear (FolderDropdownV2.tsx)**
- Toggle via `onFolderSelect(selectedFolder)` works because all 7 consumers map the prop to `setValue('folder', name === values.folder ? '' : name)`. Verified across CreditCard, Custom, Identity, Login, Note, PassPhrase, Wifi.
- `e.stopPropagation()` correctly prevents the trigger from opening the dropdown on clear-click.
- Kit usage is right (`Button` variant=`tertiary`, `Close` icon, `rawTokens.spacing6`, `aria-label`).
- minor: the right-slot wrapper uses inline `style={{ display: 'flex', ... }}` rather than a styles helper. Pre-existing code in this file already uses kit components without an extra wrapper, so a small `createStyles` entry (or a kit `Stack`) would be more consistent — non-blocking.

**Add-Item disable on required-field empty (7 modals)**
- The disabled predicates match the `Validator` schema required fields exactly:
  - title-only → CreditCard / Custom / Identity / Login / Note
  - title + passPhrase → PassPhrase
  - title + password → Wifi
- Gated on `!isEdit` so existing items stay savable; schema `validate` still catches an emptied-out title in edit mode and surfaces the inline error. That's a deliberate scope choice and matches the bug ticket wording ("Add Item button"), but worth confirming you're fine with edit mode allowing the click and showing inline errors instead of disabling.

**Other**
- minor: no tests were added/updated for the new disabled logic or the clear button. A small assertion that the save button is disabled with an empty title (and enabled once typed) on one of the modals would lock this fix in against regressions.
- minor: trailing extra blank line at the end of `FolderDropdownV2.tsx`.

Nothing blocking — Electron security posture, IPC, and persistence are untouched. Approving.

PR #475, opened 2026-05-01
[v2][Windows/MacOS/Ubuntu][Lock screen] Bugs

giorgikh93:
### Requirements

- [v2][MacOS/Linux/Windows][Lock screen] After the countdown ends on the Lock screen, there is no automatic transition to the Login screen
- [v2][MacOS/Linux/Windows][Home screen][Multiple section] The number of the selected items is displayed incorrectly
- [v2][MacOS/Linux/Windows][Home screen][Multiple section] Item's checkbox is not selected when clicking on it
- [v2][MacOS/Linux/Windows][Home screen][Multiple section] The "Item deletion" pop-up does not display the number of items being deleted
### dependencies

https://github.com/user-attachments/assets/4276922b-acb7-4361-b3d8-48fd3b6e0cee

https://github.com/user-attachments/assets/2c68a30a-f1e8-49df-bca5-f3350db323a3

Screenshot 2026-05-01 at 15 18 10

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,3 +15,7 @@ UI is built on `@tetherto/pearpass-lib-ui-kit`. All new UI — v2 redesigns of e
 **When the kit applies:** the active design is controlled by the `DESKTOP_DESIGN_VERSION` flag from `@tetherto/pearpass-lib-constants`. Currently `DESKTOP_DESIGN_VERSION === 2`, so kit components are the default for all new UI.
 
 **`V2` suffix is for coexistence only.** If a v1 file already exists for the component you're creating, add the `V2` suffix (e.g. `CardCreateMasterPasswordV2` alongside legacy `CardCreateMasterPassword`). If the component is net-new with no v1 sibling, use its natural name — **no suffix**.
+
+## React: `useEffect` dependency safety
+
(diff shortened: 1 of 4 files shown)


boris91 (approved):

giorgikh93:
@nickgr2 please review

nickgr2 (approved):
Two targeted fixes; both look correct.

**LockedScreenV2** (`useEffect` deps `[refreshMasterPasswordStatus]` → `[]`): correct. `useUserData()` returns a fresh `refreshMasterPasswordStatus` closure every render (see `node_modules/@tetherto/pearpass-lib-vault/src/hooks/useUserData.js:98`), so the original deps caused the effect to re-fire each render — which prevented the initial-load fetch from settling and likely blocked the countdown's `onFinish` navigation. Aligned with the new `CLAUDE.md` guidance.

**RecordListViewV2** (`isCheckboxSelectable={false}` on `ListItem`): correct usage of the new kit prop. Verified at `tetherto/pearpass-lib-ui-react-native-components@e2f2f7a`: `Checkbox.handleToggle` now bails out when `isCheckboxSelectable` is false, so the row's `handlePress → onSelect` is the sole toggle path and the checkbox no longer fights it.

**Kit bump** in `package-lock.json` resolves `design-system-v2` to the SHA that exposes the `isCheckboxSelectable` prop — matches.

minor: `LockedScreenV2.tsx:26` introduces trailing whitespace on the blank line after `useCountDown(...)`. Worth dropping before merge.

PR #477, opened 2026-05-01
[v2][MacOS/Linux/Windows]["Item view" mode][Identity item] There is no "Eye" icon in the "Hidden message" field

giorgikh93:
### Requirements

- [v2][Desktop]["Create master password" screen] Update password hint
- [v2][MacOS/Linux/Windows]["Item view" mode][Identity item] There is no "Eye" icon in the "Hidden message" field
- [v2][MacOS/Linux/Windows]["Item view" mode] The "Folder" field is not displayed when the item is placed into the folder
### Screenshots/Recordings

Screenshot 2026-05-01 at 16 26 51

https://github.com/user-attachments/assets/9a135004-9210-497e-a16c-893a5e06d2a8

https://github.com/user-attachments/assets/b82e1311-9d90-4830-9ae1-4abbd9558997

Screenshot 2026-05-01 at 15 39 16

### DIFF
--- a/src/containers/RecordDetails/IdentityDetailsForm/IdentityDetailsFormV2.tsx
+++ b/src/containers/RecordDetails/IdentityDetailsForm/IdentityDetailsFormV2.tsx
@@ -699,7 +699,7 @@ export const IdentityDetailsFormV2 = ({
           </MultiSlotInput>
         </div>
       )}
-      {(hasFolder || hasNote) && (
+      {(hasFolder || hasNote || hasCustomFields) && (
         <div style={styles.section}>
           <Text variant="caption">{t('Additional')}</Text>
 
--- a/src/containers/RecordDetails/IdentityDetailsForm/IdentityDetailsFormV2.tsx
+++ b/src/containers/RecordDetails/IdentityDetailsForm/IdentityDetailsFormV2.tsx
@@ -6,6 +6,7 @@ import {
   AttachmentField,
   InputField,
   MultiSlotInput,
+  PasswordField,
   Text
 } from '@tetherto/pearpass-lib-ui-kit'
 import { html } from 'htm/react'
@@ -289,13 +290,8 @@ export const IdentityDetailsFormV2 = ({
   )
   const hasAttachments = attachmentSources.length > 0
 
-  const commentValues: string[] = [
-    ...(values.note?.length ? [values.note] : []),
-    ...((values.customFields ?? []) as CustomField[])
-      .map((field) => field.note ?? '')
-      .filter(Boolean)
-  ]
-  const hasComments = commentValues.length > 0
+  const hasNote = !!values.note?.length
+  const hasCustomFields = !!values.customFields?.length
 
   const handleAttachmentPress = (attachment: Attachment) => {
     if (!attachment?.buffer || !attachment?.name) return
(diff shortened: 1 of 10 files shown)


giorgikh93:
@nickgr2 please review

nickgr2 (changes requested):
A few items worth addressing — main one is a section-structure inconsistency in the Identity form.

### `IdentityDetailsFormV2.tsx` — Hidden Messages rendered outside the "Additional" section

Previously, both `note` and custom-field notes were rendered inside a single `<div style={styles.section}>` with the `t('Additional')` caption (via the merged `commentValues` array). After this change:

- `Note` stays inside the captioned `Additional` section.
- The new `hidden-messages-multi-slot-input` block (lines 735–751) is placed **outside** that section, as a bare `MultiSlotInput` with no `styles.section` wrapper and no caption.

Two consequences:
1. If an Identity record has a Hidden Message but no Folder and no Note, the `Additional` caption disappears entirely and the Hidden Messages block renders as an unsectioned, untitled group.
2. When Folder/Note + Hidden Messages all exist, you get a captioned section followed immediately by a captionless block.

For comparison, `NoteDetailsFormV2`, `WifiDetailsFormV2`, and `CustomDetailsFormV2` keep `hidden-messages-multi-slot-input` *inside* the same Additional section as Folder/Note. Suggest doing the same here — move the `hasCustomFields` block inside the `(hasFolder || hasNote)` div and update the guard to `(hasFolder || hasNote || hasCustomFields)`.

### `CardCreateMasterPasswordV2/index.tsx` — Lingui message ID change

The string was rewritten in two ways:
- Added \"uppercase and lowercase\" (intentional content change).
- Straight apostrophe `'` changed to typographic `'` in \"aren't\".

Both alter the Lingui message ID. Confirm `npm run extract` (or equivalent) is run and existing locale catalogs are updated before release; otherwise the field will fall back to the source string in every translated locale. The typographic apostrophe in particular is easy to miss in review of the .po files.

### Minor / out of scope

- The seven copies of the `hasFolder` `MultiSlotInput`/`InputField` block are nearly identical. A small shared `<FolderReadOnlyField>` would simplify the next time this changes — fine to defer, but flagging since the duplication is now codified across all detail forms.
- The custom-field → `PasswordField` swap correctly addresses the missing eye icon. Good.

giorgikh93:
@nickgr2 addressed your point, please review again!

Regarding CardCreateMasterPasswordV2/index.tsx — Lingui message ID change
extraction and compilation of messages happening during CI, so not necessary to do it locally.

nickgr2 (changes requested):
The guard update to `(hasFolder || hasNote || hasCustomFields)` only addresses half of the feedback — the `hidden-messages-multi-slot-input` block was not moved inside the section, so the structural issue persists.

Current state at `4bc1ebf`:

```tsx
{(hasFolder || hasNote || hasCustomFields) && (
  <div style={styles.section}>
    <Text variant="caption">{t('Additional')}</Text>
    {hasFolder && (...)}
    {hasNote && (...)}
  </div>
)}

{hasCustomFields && (
  <MultiSlotInput testID="hidden-messages-multi-slot-input">
    ...
  </MultiSlotInput>
)}
```

Two problems remain:

1. **Identity with only a Hidden Message (no folder, no note):** the section now opens, renders the `Additional` caption with nothing below it, then closes — and the Hidden Messages block renders outside the section, uncaptioned. That's worse than before, where the caption simply didn't render.
2. **Identity with folder/note + Hidden Messages:** captioned section followed by a captionless `MultiSlotInput`. Same inconsistency I flagged previously.

For reference, `NoteDetailsFormV2.tsx` lines 183–199 keep the `hasCustomFields` block inside the Additional `<div style={styles.section}>` alongside `hasFolder`/`hasNote`. `IdentityDetailsFormV2` should match.

Concretely, move the `hasCustomFields` `MultiSlotInput` block inside the existing `<div style={styles.section}>` (after `hasNote`), so the final structure is:

```tsx
{(hasFolder || hasNote || hasCustomFields) && (
  <div style={styles.section}>
    <Text variant="caption">{t('Additional')}</Text>
    {hasFolder && (...)}
    {hasNote && (...)}
    {hasCustomFields && (
      <MultiSlotInput testID="hidden-messages-multi-slot-input">
        ...
      </MultiSlotInput>
    )}
  </div>
)}
```

Other items from the prior review (Lingui message-ID change confirmation, optional `FolderReadOnlyField` extraction) are unchanged and not blocking.

giorgikh93:
@nickgr2 please reveiw again! I've addressed your point

nickgr2 (approved):
Hidden Messages now renders inside the captioned `Additional` section alongside Folder and Note — structural concern from the prior review is resolved.

src/containers/RecordDetails/IdentityDetailsForm/IdentityDetailsFormV2.tsx
```diff
@@ -702,26 +699,54 @@ export const IdentityDetailsFormV2 = ({
...
       )}
-
-      {hasComments && (
+      {(hasFolder || hasNote || hasCustomFields) && (
         <div style={styles.section}>
           <Text variant="caption">{t('Additional')}</Text>
 
-          <MultiSlotInput testID="comments-multi-slot-input">
-            {commentValues.map((comment, index) => (
+          {hasFolder && (
```
boris91: i see this part being used multiple times, can we have separate component for it to be shared among those places? 🙏

boris91 (dismissed):

boris91 (approved):

PR #482, opened 2026-05-04
Lower app size for users

denislupookov:
### Changes

- Get rid of unneeded prebuild folders from holepunch side. Libs like bare-sidecar, rocksdb-native, bare-tls ship /prebuilds with binaries for every platform (darwin/linux/win32/android/ios × x64/arm64). In afterPack we walk the bundled node_modules and keep only the dirs matching the build target.
- Get rid of react native deps

Analytics on my mac arm:

Initial	                        1.1 GB (/out folder)	336 MB (dmg)
Prebuild prune only	580 MB	                        178 MB
Both	                        478 MB                             171 MB

---
  - https://app.asana.com/0/0/1213987034629408

### DIFF
--- a/scripts/afterPack.cjs
+++ b/scripts/afterPack.cjs
@@ -1,21 +1,151 @@
 #!/usr/bin/env node
 /**
- * Electron-builder afterPack hook for Linux:
+ * Electron-builder afterPack hook.
  *
- * 1. Remove chrome-sandbox — squashfs (AppImage) cannot preserve SUID bits,
- *    so Chromium's setuid sandbox probe crashes before Node.js even starts.
+ * 1. Prune native `prebuilds/<platform>-<arch>` dirs that don't match the
+ *    current target. Native modules in the bare/holepunch ecosystem ship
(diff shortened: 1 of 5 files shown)


boris91 (dismissed):

torkinos (approved):

scripts/afterPack.cjs
```diff
@@ -1,21 +1,151 @@
...
+  }
+}
+
+async function prunePrebuilds(context) {
+  const { appOutDir, electronPlatformName, arch, packager } = context
+  const archName = ARCH_NAMES[arch] ?? String(arch)
+  const target = `${electronPlatformName}-${archName}`
+
+  let appRoot
+  if (electronPlatformName === 'darwin' || electronPlatformName === 'mas') {
```
hakaari, line 44: `mac` or `mac` @denislupookov ?

denislupookov: it is actually mac app store, specifically for apple store, I just left it for future. If needed I will remove @hakaari

denislupookov: https://www.electron.build/app-builder-lib.typealias.electronplatformname

PR #497, opened 2026-05-07
2FA: Standalone 2FA Form

giorgikh93:
### Changes

- OTP remove approach without deleting the login record
### Testing Notes

- Tested locally
### Screenshots/Recordings

https://github.com/user-attachments/assets/b6d3c593-a1f9-48d6-aa9e-7bc3c7abfef0

https://github.com/user-attachments/assets/aeceee59-b6d7-49f6-9c40-71936caac205

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,5 @@ appling/node_modules/
 test-results
 
 out/
+
+src/PearPass
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,6 +1,7 @@
 import { eslintConfig } from '@tetherto/tether-dev-docs'
 
 export default [
+  { ignores: ['src/PearPass/**'] },
   ...eslintConfig,
   {
     files: ['**/*.{js,jsx,mjs,cjs}'],
--- a/lingui.config.js
+++ b/lingui.config.js
@@ -8,7 +8,10 @@ export default defineConfig({
     {
       path: './src/locales/{locale}/messages',
       include: ['./src'],
-      exclude: ['./src/native-messaging-bridge/node_modules/**']
+      exclude: [
+        './src/native-messaging-bridge/node_modules/**',
+        './src/PearPass/**'
+      ]
     }
   ],
   format: formatter({ style: 'minimal' }),
--- a/package-lock.json
+++ b/package-lock.json
@@ -5912,7 +5912,7 @@
     "node_modules/@tetherto/pearpass-lib-vault": {
       "name": "pearpass-lib-vault",
       "version": "0.0.85",
-      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-vault.git#21d29ed5ba6c2c1272df6017389c0192f3178581",
+      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-vault.git#538be529307326322ef02c672d272bdf9137fb88",
       "license": "Apache-2.0",
       "dependencies": {
         "@reduxjs/toolkit": "2.5.0",
--- a/src/app/Routes/index.js
+++ b/src/app/Routes/index.js
@@ -1,5 +1,5 @@
 import { AUTHENTICATOR_ENABLED } from '@tetherto/pearpass-lib-constants'
-import { OtpRefreshProvider } from '@tetherto/pearpass-lib-vault'
+import { OtpRefreshProvider, RECORD_TYPES } from '@tetherto/pearpass-lib-vault'
 import { html } from 'htm/react'
 
 import { LayoutWithSidebar } from '../../containers/LayoutWithSidebar'
@@ -66,7 +66,7 @@ export const Routes = ({
 
   if (currentPage === 'vault') {
     const isAuthenticator =
-      AUTHENTICATOR_ENABLED && data?.recordType === 'authenticator'
+      AUTHENTICATOR_ENABLED && data?.recordType === RECORD_TYPES.OTP
 
     const VersionBasedMainView = isV2() ? MainViewV2 : MainView
 
--- a/src/containers/AppHeaderContainer/AppHeaderContainer.tsx
+++ b/src/containers/AppHeaderContainer/AppHeaderContainer.tsx
@@ -47,13 +47,6 @@ export const AppHeaderContainer = () => {
     return null
   }
 
-  if (
-    AUTHENTICATOR_ENABLED &&
-    routerData?.recordType === 'authenticator'
-  ) {
-    return null
-  }
-
   const isFavoritesView = isFavorite(routerData?.folder ?? '')
   const selectedFolder =
     routerData?.folder && !isFavoritesView ? routerData.folder : undefined
--- a/src/containers/AppHeaderContainer/AppHeaderContainer.tsx
+++ b/src/containers/AppHeaderContainer/AppHeaderContainer.tsx
@@ -1,5 +1,3 @@
-import React from 'react'
-
 import { AUTHENTICATOR_ENABLED } from '@tetherto/pearpass-lib-constants'
 import { ContextMenu, NavbarListItem, useTheme } from '@tetherto/pearpass-lib-ui-kit'
 import {
@@ -72,7 +70,7 @@ export const AppHeaderContainer = () => {
     { type: 'password',                 label: t('Password'),          icon: <Key color={iconColor} /> },
     { type: RECORD_TYPES.CUSTOM,        label: t('Other'),             icon: <GridView color={iconColor} /> },
     ...(AUTHENTICATOR_ENABLED
-      ? [{ type: 'authenticator', label: t('Authenticator Code'), icon: <QrCode color={iconColor} /> }]
+      ? [{ type: RECORD_TYPES.OTP, label: t('Authenticator Code'), icon: <QrCode color={iconColor} /> }]
       : [])
   ]
 
--- a/src/containers/Modal/CreateOrEditCategoryWrapper/CreateOrEditAuthenticatorModalContent/CreateOrEditAuthenticatorModalContent.styles.ts
+++ b/src/containers/Modal/CreateOrEditCategoryWrapper/CreateOrEditAuthenticatorModalContent/CreateOrEditAuthenticatorModalContent.styles.ts
@@ -1,0 +1,13 @@
+import { rawTokens } from '@tetherto/pearpass-lib-ui-kit'
+
+export const createStyles = () => ({
+  form: {
+    display: 'flex' as const,
+    flexDirection: 'column' as const,
+    gap: `${rawTokens.spacing8}px`,
+    width: '100%'
+  },
+  sectionLabel: {
+    marginTop: `${rawTokens.spacing8}px`
+  }
+})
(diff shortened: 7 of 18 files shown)


giorgikh93:
@nickgr2 please review

nickgr2 (changes requested):
A few issues — the form as-is doesn't produce records the Authenticator view can show.

### Blocker — wrong record type on submit

`CreateOrEditAuthenticatorModalContent.tsx`, in `onSubmit`:

```tsx
const data = {
  type: RECORD_TYPES.LOGIN,   // <- looks like a copy from CreateOrEditLoginModalContentV2
  ...
}
```

The wrapper routes `RECORD_TYPES.OTP` to this component, and `AuthenticatorView`'s "Add Code" CTA calls `handleCreateOrEditRecord({ recordType: RECORD_TYPES.OTP })`. With `type: RECORD_TYPES.LOGIN`, the saved record is created as a Login. `AuthenticatorView` filters with `(records || []).filter((r) => r.otpPublic)` (and `useRecords({ filters: { hasOtp: true } })`), so the new record won't appear in the view it was created from. Should be `RECORD_TYPES.OTP`.

### Blocker — no OTP secret field

The form only captures `title`, `note`, and `attachments`. There's no input for the actual TOTP/HOTP secret (or issuer / period / digits / algorithm / QR scan). `CreateOrEditLoginModalContentV2` reads `formValues.otpSecret` and stores it as `data.otpInput`, which the vault then turns into `otpPublic`. Without an equivalent here, the saved record won't have `otpPublic` and won't show up in `AuthenticatorView` even if the type bug above is fixed.

If this PR is intentionally a UI-only scaffold and the secret-input + OTP plumbing is coming in a follow-up, please call that out in the description and link the follow-up issue — right now the end-to-end flow is non-functional and that's hard to evaluate from an empty PR body.

### PR description is empty

Only the template headings remain. Per `CONTRIBUTING.md`, please fill in at least `Changes` and `Testing Notes` so reviewers can confirm scope vs. intent (especially given the partial behavior above).

### Other notes

- `onTypeChange` is required in the prop type but renamed to `_onTypeChange` and unused. Matches the sibling V2 forms, so fine — just flagging.
- `MultiSlotInput` wrapping a single `InputField` for the Comment field is unusual compared to how it's used in `LoginV2` (slots a real OTP/recovery group). Consider whether the bare field is enough.

`testID` / `data-testid` conventions and kit-component usage look right; UI-kit compliance is fine.

giorgikh93:
@nickgr2  please review again

nickgr2 (changes requested):
The previous blockers are addressed — `PasswordField` for the secret is in place, and storing as `RECORD_TYPES.LOGIN` with `otpInput` is the correct shape (the vault's `validateAndPrepareRecord` only accepts the existing record types, and `AuthenticatorView` filters via `hasOtp` on login records). Confirmed against `pearpass-lib-vault` at the new lock SHA `538be529...`.

A few new issues introduced in the *Data Synchronization & Mapping* commit:

### Two search bars stacked in authenticator view

`AppHeaderContainer.tsx` deleted the early `return null` for the authenticator route, so on the OTP route it now renders alongside `AuthenticatorView`'s own `<InputSearch>`. They have independent state/scope (AppHeader's `SearchField` reads from `useAppHeaderContext`, AuthenticatorView's `InputSearch` is local with its own `searchValue`). The standard V2 pattern (see `MainViewV2.tsx`) is *AppHeader has the search, the page reads it from context*. Either restore the early return for the OTP route, or switch `AuthenticatorView` to the V2 pattern (drop its local `InputSearch`, read `searchValue` from `useAppHeaderContext`, push the existing `searchPattern` filter from there).

### `handleStripOtp` is fire-and-forget

`useRecordActionItems.js:51-60`:
```js
const handleStripOtp = () => {
  ...
  updateRecords([updatedRecord])      // returns Promise<void>, not awaited
  if (...) navigate(...)
  closeModal?.()                      // runs before the dispatch resolves
}
```

`updateRecords` is async. Without `await` and without `onError`, a failed update closes the modal silently and leaves stale state. The existing delete path uses `useRecords({ onCompleted: closeModal })` precisely so the modal stays open with `isLoading` until the op completes. Make this `async`, await `updateRecords`, and pass an error handler that surfaces a toast (matches the pattern in `CreateOrEditAuthenticatorModalContent.onError`).

### Delete dialog copy is misleading on the strip-OTP path

When `onConfirm` is supplied, `DeleteRecordsModalContentV2` still renders title `"Delete 1 Item"`, confirm text `"Are you sure to delete the selected item?"`, and button label `"Delete Item"` — but the action keeps the underlying login record and only strips OTP fields. From the user's perspective they pressed Delete and the record disappeared, which works *if* the record was OTP-only; for a login that also had a password they'd reasonably believe the whole entry was deleted. Either accept overrides for `dialogTitle` / `confirmText` / `submitLabel` from props (e.g. *"Remove Authenticator Code"* / *"This will keep the underlying login but remove the 2FA code"* / *"Remove Code"*), or use a different confirmation modal for this flow.

### Edit from authenticator view opens the full Login form

`useRecordActionItems.handleEdit` keys the modal off `record?.type`, which for OTP records is `LOGIN`, so Edit-from-authenticator-view routes to `CreateOrEditLoginModalContentV2` (with username, password, passkey, websites, custom fields, etc.) rather than the simpler authenticator form Create just used. The hook now has `recordType` / `isOtpContext` available — consider routing to `CreateOrEditAuthenticatorModalContent` when in OTP context. Functionally LoginV2 does handle OTP correctly, so this is more a UX consistency point than a correctness blocker.

### `recordType` typing on `useRecordActionItems`

`useRecordActionItems.d.ts` widens to `recordType?: string`. Since callers pass `RECORD_TYPES.OTP` and the hook only branches on that one value, narrowing to `RECORD_TYPES['OTP']` (or a union of the enum values you care about) prevents typo drift.

### PR description still empty

Carrying this over from the prior review — only the template headings remain. Per `CONTRIBUTING.md`, please fill in `Changes` / `Testing Notes` so the scope is reviewable from the PR alone.

giorgikh93:
@nickgr2 addressed your points, please reveiw

nickgr2 (changes requested):
Most of the prior CR points are resolved — the search bar collision is fixed by switching `AuthenticatorView` to the V2 pattern, `handleStripOtp` is now `async` with a `try/catch` toast, Edit-from-OTP routes through `RECORD_TYPES.OTP` to the authenticator modal, and the `recordType` typing on `useRecordActionItems` is narrowed.

Two carry-overs and one new concern:

### Carry-over — delete dialog copy still misleading on the strip-OTP path

`DeleteRecordsModalContentV2.tsx` was not touched in `e8b35de`. When `onConfirm` is the OTP-strip handler, the dialog still renders title `"Delete 1 Item"`, body `"Are you sure to delete the selected item?"`, and button `"Delete Item"` — but pressing it keeps the underlying login and only strips `otpInput` / `otp` / `otpPublic`. A user with a login that has both a password and an authenticator code would reasonably believe Delete removed the whole record. Please accept overrides for `dialogTitle` / `confirmText` / `submitLabel` (e.g. *"Remove Authenticator Code"* / *"This will keep the underlying login but remove the 2FA code"* / *"Remove Code"*) or branch to a dedicated confirmation modal for this flow.

### Out of scope — `src/PearPass` ignore patterns

`20bc560` adds `src/PearPass` to `.gitignore`, `eslint.config.js` (`{ ignores: ['src/PearPass/**'] }`), `lingui.config.js` exclude, and `tsconfig.json` exclude. No `src/PearPass` directory exists in the repo on this branch, and nothing in the 2FA feature references it. This looks like a local-checkout artifact (a sibling clone or symlink in your working tree) that doesn't belong in this PR — please drop these four hunks or split them off into their own PR with the context for why the rest of the codebase needs to ignore that path.

### Carry-over — PR description still empty

Only template headings remain. Per `CONTRIBUTING.md`, please fill in `Changes` / `Testing Notes` so the scope is reviewable from the PR alone.

### Minor nits (non-blocking)

- `useRecordActionItems.d.ts` now types `recordType?: 'login' | 'otp'` as string literals. Since the hook only branches on `RECORD_TYPES.OTP`, narrowing to `typeof RECORD_TYPES['OTP'] | typeof RECORD_TYPES['LOGIN']` (or just the OTP one — the LOGIN branch is the default) would keep this in sync with the enum if values change.
- `.gitignore` and `tsconfig.json` both end without a trailing newline — pre-existing on the file, but the added line preserves the issue. Worth adding a newline while you're already touching these.

giorgikh93:
@nickgr2 
copy stays the same for now "Delete 1 Item", body "Are you sure to delete the selected item?", this is generic texts and we are okay with that for now.

added PR changes and Tested notes, please reveiw again

nickgr2:
> copy stays the same for now ... this is generic texts and we are okay with that for now

Noted — your call on that one.

> added PR changes and Tested notes, please reveiw again

`Changes` / `Testing Notes` are in. The head SHA hasn't moved since [the review at 09:01](https://github.com/tetherto/pearpass-app-desktop/pull/497#pullrequestreview-3527437791) (`20bc560`), so there's no new code to look at. The other carry-over from that review — the four `src/PearPass` ignore hunks across `.gitignore`, `eslint.config.js`, `lingui.config.js`, and `tsconfig.json` — is still open. Drop those (or split them into a separate PR) and I'll take another pass once you push.

giorgikh93:
@nickgr2 src/PearPass/by-arch/ contains the bundled Electron app with a full node_modules inside src/, and ESLint is scanning all of it. Needed to add it to ignores in the flat config. I haven't pushed new changes, you can review again and approve.

nickgr2:
> src/PearPass/by-arch/ contains the bundled Electron app with a full node_modules inside src/, and ESLint is scanning all of it. Needed to add it to ignores in the flat config.

Fair — withdrawing the out-of-scope flag on the `src/PearPass` ignores. `e8b35de` adds the same exclusion to `tsconfig.json` and `lingui.config.js` alongside the ESLint config, so the four hunks are consistent toolchain housekeeping for a generated/vendored tree, and CI lint passes at `20bc560`. Leave them in.

The dialog-copy carry-over is still the open blocker — `DeleteRecordsModalContentV2.tsx` wasn't touched in `e8b35de`, so when `onConfirm` is the OTP-strip handler the dialog still says *"Delete 1 Item"* / *"Are you sure to delete the selected item?"* / *"Delete Item"* while the action actually keeps the underlying login. Once that's pushed (override props for `dialogTitle` / `confirmText` / `submitLabel`, or branch to a dedicated modal), I can re-review.

giorgikh93:
@nickgr2 I didn't touch to dialog messages as they are generic messages and suit well to the current scenario. You can skip this and re-review

nickgr2:
Reasonable on the OTP-only case — 'Delete Item' from authenticator view reads cleanly there since the entry does disappear from the user's view.

One edge that remains, just for the record: login+OTP records (login with password + OTP added via the LoginV2 form) viewed in AuthenticatorView still route through `handleStripOtp` (because `useRecordActionItems` flags any login record as `isAuthenticatorLoginRecord` whenever `isOtpContext` is true). That path keeps the password, so for those records 'Delete Item' actually preserves the underlying record. Non-blocking — UX call is yours.

No new commits since 20bc560 so my prior review state stays on record. Push anything (even a trivial commit) when you want a fresh pass with the dialog-copy point dropped.

giorgikh93:
@nickgr2 I have nothing to push more, you can approve or tell me if I need to adjust anything

nickgr2:
Understood. With the PR description now filled in and the other points (dialog copy, typing/newline nits) already tagged non-blocking in the thread, there's nothing more for you to fix from my side.

The formal CHANGES_REQUESTED on 20bc560 stays on record — I can't toggle a review without a new SHA — but treat the substantive feedback as resolved. Heads up that nickgr2 isn't a collaborator on this repo regardless, so an approval here wouldn't unblock branch protection on its own. The maintainer will need to merge.

boris91 (approved):

torkinos (approved):

PR #499, opened 2026-05-11
[v2] [Desktop] Redesign authenticator details screen , [v2] [Desktop] Redesign authenticator screen

nikushaomanidze:
### Changes

[v2] [Desktop] Redesign authenticator details screen
[v2] [Desktop]Redesign authenticator details screen

### Testing Notes

Mac od Desktop app

### Screenshots/Recordings

https://github.com/user-attachments/assets/bb4ecd29-7e20-488e-9648-6b7619524110


torkinos:
@nickgr2 please review

boris91 (dismissed):

nickgr2 (commented):
Solid migration to the kit — converts `AuthenticatorView` to `Breadcrumb` + `ListItem` + `ContextMenu`, introduces `OtpCodeFieldV2`, and adopts the same grouped-section pattern as `MainViewV2`. A few findings below.

**Missing tests for new behavior** — `OtpCodeFieldV2` has no colocated test (the v1 `OtpCodeField` does). The rewritten `AuthenticatorView` adds multi-select toggling, Escape-to-exit, sort menu, and favorite/move/delete bulk actions — none of which have unit or integration coverage, and there's no `*.test.*` file alongside `pages/AuthenticatorView/`. Per `REVIEW_GUIDELINES.md` §10 these are behavior changes that warrant tests at the component level (mirroring `RecordListViewV2.test.tsx` for `MainViewV2`).

**Type-assertion escape hatch on `Text` style** — in `src/components/OtpCodeFieldV2/index.tsx`:

```tsx
<Text
  variant="caption"
  style={
    { color: timerColor, minWidth: 28, textAlign: 'right' } as object
  }
>
```

The `as object` cast bypasses the kit's restricted `style` typing (the kit uses `react-strict-dom`, which is intentionally narrow). Per `REVIEW_GUIDELINES.md` §6, prefer adding these to `createStyles` and passing a named style key, or use a typed cast (e.g. `as React.CSSProperties` if that's what the prop allows). The `color` is already supported as a first-class prop on `Text` (`color={timerColor}`), so only `minWidth`/`textAlign` need to live in styles.

**Duplicate `rowRightElement` style** — `src/pages/AuthenticatorView/styles.ts` defines `rowRightElement` identically to the one already exported from `src/containers/RecordListView/RecordListViewV2.styles.ts` (which the file imports as `listStyles`). Drop the local copy and use `listStyles.rowRightElement` to avoid drift.

**Inline override of `sectionHeader`** — in `pages/AuthenticatorView/index.js`, both group headers do:

```js
style=${{ ...listStyles.sectionHeader, color: theme.colors.colorTextPrimary, cursor: 'default' }}
```

The shared `sectionHeader` is built for the collapsible/clickable list pattern (cursor pointer, secondary color). Adding a `staticSectionHeader` variant to `RecordListViewV2.styles` and using that here would make the static, non-interactive case explicit rather than relying on per-call overrides. Minor.

**Hardcoded sizing in OtpCodeFieldV2** — `gap: 2`, `height: 6` (timer track), and `minWidth: 28` (timer label) in `styles.ts` / inline. Not blocking, but if there's a `rawTokens.spacing2` / similar, prefer that for consistency.

**Pre-existing carryovers (not blockers)**
- `isCheckboxSelectable={false}` on `ListItem` — same pattern as `RecordListViewV2.tsx`; the prop doesn't exist on `ListItem` (see `node_modules/@tetherto/pearpass-lib-ui-kit/src/components/ListItem/ListItem.tsx`) and is silently ignored. Worth removing here and there in a follow-up.
- `useRecords({ variables: { filters: { hasOtp: true, ...` — `hasOtp` is not forwarded to the selector (only `searchPattern, type, folder, isFavorite` are), so the `.filter((r) => r.otpPublic)` client-side filter is doing the real work. Pre-existing from v1; flagging only because the dead arg was carried over.

**Verified**
- All vault and kit imports exist (`Breadcrumb`, `ContextMenu`, `NavbarListItem`, `ListItem`, `Title`, `Text`, `Button`, `useTheme`, all icons, `useFolders`, `useRecords.updateFavoriteState`).
- Naming follows the V2 rule: `OtpCodeFieldV2` because `src/components/OtpCodeField/` is the v1 sibling; `AuthenticatorView/index.js` correctly stays unsuffixed (no v1 sibling — same file rewritten).
- All user-facing strings go through Lingui.
- Removing the `routerData?.recordType === 'authenticator' → return null` in `AppHeaderContainer.tsx` is consistent with the new design: the global header (which drives `useAppHeaderContext().searchValue` consumed by this view) now stays visible above the authenticator listing/details.
- No Electron-security, IPC, `safeStorage`, `shell.openExternal`, or auto-updater changes; no new dependencies; no new SVGs under `src/` (uses pre-existing `src/svgs/ItemCardIllustration`).

nikushaomanidze:
@nickgr2 please review again

nickgr2 (approved):
Prior findings are addressed — tests added for `OtpCodeFieldV2` and `AuthenticatorView`, `as object` cast on `Text` style is gone, the duplicate `rowRightElement` now reuses `listStyles.rowRightElement`, and the inline `sectionHeader` override is replaced by the new shared `listStyles.staticSectionHeader`. Approving.

A couple of minor observations on the new code, none blocking:

**Inconsistent `RECORD_TYPES.LOGIN` vs literal `'login'`** — `RECORD_TYPES` is imported at the top of `pages/AuthenticatorView/index.js` but the empty-state Add Code button hardcodes the string:

```js
onClick=${() => handleCreateOrEditRecord({ recordType: 'login' })}
```

Prefer `RECORD_TYPES.LOGIN` for consistency with the rest of the file (and the rest of the codebase routing through `CreateOrEditCategoryWrapper`, which compares against `RECORD_TYPES.LOGIN`).

**Pre-existing carryover: `RECORD_TYPES.OTP` is undefined** — `handleRecordPress` does:

```js
navigate('vault', {
  recordId: isAlreadyOpen ? '' : record.id,
  recordType: RECORD_TYPES.OTP
})
```

`@tetherto/pearpass-lib-vault/src/constants/recordTypes.js` only freezes `NOTE | CREDIT_CARD | CUSTOM | IDENTITY | LOGIN | WIFI_PASSWORD | PASS_PHRASE` — there is no `OTP` member, so this passes `undefined` as `recordType` to the router. This was already true in the v1 file (`handleRecordClick`) so it's a carryover, not a regression. Worth a follow-up to either add `OTP` (or `AUTHENTICATOR`) to the enum upstream or drop the field if the router never reads it for this view.

**Verified**
- All kit imports exist: `Breadcrumb`, `Button`, `ContextMenu`, `ListItem`, `NavbarListItem`, `Text`, `Title`, `useTheme`, `rawTokens`, and icons `Add`, `CalendarToday`, `Check`, `Checklist`, `ContentCopy`, `FilterList`, `ImportExport`, `SortByAlpha` (all present under `node_modules/@tetherto/pearpass-lib-ui-kit/src/icons/components/`).
- Vault imports: `useOtp`, `useTimerAnimation`, `formatOtpCode`, `OTP_TYPE`, `useFolders`, `useRecords`, `groupOtpRecords`, `isExpiring`, `RECORD_TYPES` all exported from `pearpass-lib-vault/src/index.js`. `useRecords` returns `updateFavoriteState` as used.
- `MultiSelectActionsBar` prop shape matches the call site (`selectedCount`, `allSelectedFavorited`, `canMove`, `onMove`, `onToggleFavorite`, `onDelete`).
- `SettingsItemKey.ImportItems` exists in `SettingsViewV2.tsx`.
- The deep import `from '@tetherto/pearpass-lib-vault/src/types'` for `OtpPublic` mirrors the v1 `OtpCodeField` pattern — not a new escape hatch introduced here.
- New `staticSectionHeader` in `RecordListViewV2.styles.ts` cleanly replaces the per-call `cursor:default` override; reusable from the main list view too if it ever needs a non-collapsible mode.
- `LoginRecordDetailsFormV2.tsx` swap from `OtpCodeField` → `OtpCodeFieldV2` puts the OTP slot inside the existing credentials `MultiSlotInput` (matching the new grouped design via `isGrouped`); the old standalone `otp-multi-slot-input` block is removed cleanly.
- All user-facing strings go through Lingui (`i18n._()`).
- `testID` / `data-testid` present on all new interactive controls (multi-select toggle, sort menu trigger, sort options, copy buttons, empty-state buttons, record rows).
- Removing the authenticator-vault `return null` in `AppHeaderContainer.tsx` and updating its test to assert the header now renders is consistent with the new design (the breadcrumb + header actions live inside `AuthenticatorView`, but the global header still drives the search box this view consumes).

torkinos (approved):

boris91 (approved):

PR #503, opened 2026-05-13
ci:adding workflows authorize gate + PR-head context

kapildev421:
### Requirements

- Run a CI workflow (lint, unit tests) for every PR opened against main such that:
- The workflow validates the PR head code, not the target branch.
- Triggers on opened, reopened, and synchronize.
- Runs only for trusted contributors (org members, collaborators, or external PRs explicitly opted-in by a maintainer via the verify label) — required because this is a public repo.
- using scherermichael-oss/action-has-permission to check the permission

### Changes

- Replaces lint.yml + test.yml with a three-file CI pipeline

### Testing Notes

- Open a fresh test PR with a small no-op change against main.
- In the Actions tab, confirm a workflow run named "PR CI" appears within ~10s of opening.
- Open the run → authorize job → "Check authorization" step. Confirm a ::notice:: line of the form Same-repo PR — authorized (for a branch on this repo) or Trusted author (MEMBER) — authorized (for a fork PR from an org member).
- Open ci / quality job → "Checkout code" step. Confirm the HEAD is now at <SHA> line matches the "Latest commit" SHA shown at the top of the PR in the GitHub UI. This proves the workflow ran against the PR head, not main.
- Confirm Install dependencies, Run ESLint, Run tests all complete green.
- Trigger synchronize: push another commit (or git commit --allow-empty -m "trigger sync" && git push) to the PR branch → confirm a brand-new "PR CI" run starts automatically, and the previous run is cancelled by the concurrency group.
- Trigger reopened: close the PR, then reopen it → confirm a brand-new "PR CI" run starts.

- No secrets passed to the reusable — ci-pr.yml does not declare a secrets: block, so even if the PR head contained malicious test code, it has no token to exfiltrate.

### Screenshots/Recordings
N/A- only github Actions changes

### dependencies

- No dependent PRs in this repo


torkinos:
@nickgr2 please review

nickgr2 (approved):
Solid security work overall. SHA-pinning, `persist-credentials: false`, no `secrets:` block on the reusable, and the label-strip-on-synchronize that forces re-review per commit are all the right calls. A few notes:

**`authorize-pr/action.yml` — strip-label log is unconditional**

```yaml
gh api -X DELETE \
  "repos/${{ github.repository }}/issues/${PR_NUMBER}/labels/${LABEL}" \
  2>/dev/null || true
echo "::warning::Removed '${LABEL}' label — new commits from non-writer. Re-review required."
```

The `2>/dev/null || true` swallows failures (404 because the label was already gone, permissions issue, rate limit). The follow-up `::warning::` claims removal happened regardless. Suggest:

```yaml
if gh api -X DELETE \
  "repos/${{ github.repository }}/issues/${PR_NUMBER}/labels/${LABEL}" 2>/dev/null; then
  echo "::warning::Removed '${LABEL}' label — new commits from non-writer. Re-review required."
else
  echo "::warning::Tried to remove '${LABEL}' label (may already be absent)."
fi
```

**`authorize-pr/action.yml` — unmaintained third-party action**

`scherermichael-oss/action-has-permission` is unmaintained (last release 2019, no commits in years). SHA-pinning makes the current run safe, but it's a tripwire for future maintainers: if the upstream repo gets transferred/deleted, that pinned action will start failing and somebody will have to scramble. The same permission check is a one-line shell using the existing token:

```bash
PERM=$(gh api "repos/${BASE_REPO}/collaborators/${GITHUB_ACTOR}/permission" --jq '.permission' 2>/dev/null || echo "none")
case "$PERM" in admin|maintain|write) HAS_WRITE=1 ;; *) HAS_WRITE=0 ;; esac
```

Not a blocker — pinning is sufficient — but worth replacing while the change is fresh.

**minor: `ci-push.yml` branches syntax**

```yaml
on:
  push:
    branches: main
```

Other workflows in this PR use list form (`branches: [main]`). Both are valid YAML, but stay consistent within the same change.

**minor: `labeled` trigger fires on every label**

`types: [opened, reopened, synchronize, labeled]` runs the authorize job for every label add (e.g. `bug`, `enhancement`, `area/ui`), not just `verify`. The job correctly no-ops in those cases, but it's a few seconds of wasted CI per labeling action. Optional optimization:

```yaml
jobs:
  authorize:
    if: github.event.action != 'labeled' || github.event.label.name == 'verify'
```

**Question — `branches: [main]` on `ci-pr.yml`**

The old `lint.yml` / `test.yml` ran on `main` AND `dev`. The new `ci-pr.yml` only filters `main`. Intentional drop of `dev`, or is `dev` no longer a thing in this repo?

Aside from those, the design looks right: same-repo trusted, write-permission actor trusted, fork PRs gated behind the `verify` label, label stripped on subsequent pushes from non-writers to force re-review, no secrets exposed to the checked-out fork code. APPROVE.

kapildev421:
@nickgr2 please review

torkinos (approved):

hakaari (approved):

PR #504, opened 2026-05-13
Add ability to close details view

ndolidzee:
### Requirements

Allow users to close the Detail Item view.

### Changes

Update ui-kit to use new KeyboardTab icon
Add ability to close details view
Add min-width to window.

### Screenshots/Recordings

https://github.com/user-attachments/assets/98f238c7-4782-4640-8b89-ea32df35f0ba

### dependencies

---
  - https://app.asana.com/0/0/1214440446024239

### DIFF
--- a/electron/main.cjs
+++ b/electron/main.cjs
@@ -504,6 +504,7 @@ function createWindow() {
   mainWindow = new BrowserWindow({
     width: 1440,
     height: 1024,
+    minWidth: 816,
     ...(isMac && isV2
       ? {
           titleBarStyle: 'hidden',
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "@tetherto/pearpass-lib-data-export": "git+https://github.com/tetherto/pearpass-lib-data-export.git",
         "@tetherto/pearpass-lib-data-import": "git+https://github.com/tetherto/pearpass-lib-data-import.git",
         "@tetherto/pearpass-lib-native-messaging-bridge": "git+https://github.com/tetherto/pearpass-lib-native-messaging-bridge.git",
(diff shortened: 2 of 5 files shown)


torkinos:
@nickgr2 please review

nickgr2 (approved):
LGTM. Close-button wiring is clean — reusing the existing `handleCollapse` (already called from the `useEffect` for stale records) avoids a parallel exit path. Window/Pear `minWidth: 816` matches the layout floor (sidebar + main 300 + side 250 ≈ 800), so the new sideview min-widths can't push the chrome below its own min.

A couple of minor nits, none blocking:

- **`minor:` `KeyboardTab` for a close-detail action.** The icon reads as "indent / push-right" rather than "close". For a panel collapse where the panel sits on the right, that direction actually works — but consider whether a `Close` / `CloseFullscreen` glyph would be more discoverable. Not blocking.
- **`minor:` Hardcoded split-pane widths.** `flex: '0 1 350px'`, `minWidth: '300px'`, and `minWidth: '250px'` in `LayoutWithSidebar/index.js` are inline literals rather than tokens. These are layout primitives (not design-system spacing) so it's reasonable, but worth flagging in case the kit grows a `LayoutSplit` primitive — keeping them centralized would prevent them drifting out of sync with the chrome `minWidth: 816`.
- **`minor:` Unrelated lockfile churn.** `package-lock.json` flips `fsevents` to `dev: true` and removes the `@tetherto/swarmconf` optional stub. Looks like incidental `npm install` output rather than anything intentional — fine to leave, but worth a glance to confirm it isn't a sign the lockfile was regenerated against a different npm version.
- **`minor:` Re-rendering.** The `style={{ display: 'flex', alignItems: 'center', gap: ... }}` object in `headerActions` is recreated every render. The header is small so impact is negligible — only worth hoisting if `ItemScreenHeader` starts memoizing.

torkinos (approved):

boris91 (approved):

PR #507, opened 2026-05-14
Added matching logic for 2fa

denislupookov:
### Changes
Added matching logic for 2fa

### Screenshots/Recordings

https://github.com/user-attachments/assets/b45968d8-34d8-4c1b-bcbf-352aacc97993


torkinos:
@nickgr2 please review

boris91 (approved):

nickgr2 (commented):
A few observations from the diff review against head SHA `9f898907`:

**Question — `package.json` ui-kit reference**

The diff swaps `@tetherto/pearpass-lib-ui-kit` from a pinned commit (`#52a0203f...`) to a branch reference (`#design-system-v2`). The lockfile still pins to a specific tip (`#ce3ab5de...`), so `npm ci` is reproducible — but any future `npm install` of an unrelated package can re-resolve to a newer branch tip and silently shift the ui-kit. The other pinned dep on the same line (`pearpass-lib-vault-core`) keeps its commit hash. Is this branch ref intentional for the duration of design-system-v2 work, with a plan to pin once the branch lands?

**`src/hooks/useRecordActionItems.d.ts` — indentation regression**

The diff drops `recordType?` but the next line gets left-aligned at column 0:

```ts
  >
onSelect?: () => void   // <-- should be indented
  onClose?: () => void
```

Confirmed in the working tree at this SHA. Re-indent to two spaces to match the surrounding type.

**`CreateOrEditAuthenticatorModalContent.tsx` — minor concerns**

- L107 `parsedEmail` — `parseOtpInput` returns whatever follows the first `:` in the URI label (`'GitHub:user@example.com'` → `'user@example.com'`, but for many providers the label is a display name like `'Account 1'`, not an email). On the create path (L140) you persist that value into `data.username`. Worth either renaming the variable (`parsedLabel`) and/or only assigning when it looks like an email, otherwise the username field can end up with a display-name string the user didn't enter.
- Combobox `onClear` (L234) clears `linkedRecordId` only. The title field re-enables but still holds the linked record's title (set in `onSelect`, L253) — the originally typed title isn't restored. Minor, but a user who links by accident loses what they typed.
- The styles file still exports `sectionLabel`, no longer referenced after the "Additional" / Comment / Attachments block was removed. Unused.

**`CreateOrEditLoginModalContentV2.tsx`**

Moving the OTP error from `PasswordField.error` to `MultiSlotInput.errorMessage` is fine, but verify visually that long error strings (`'Unsupported OTP type'`, `'OTP URI has an invalid secret'`) render correctly under the slot — the kit's `errorMessage` is documented as "displayed beneath the grouped container" and historically inline `error` and slot-level `errorMessage` differ in spacing.

**i18n note on validator messages**

`validateOtpInput` (lib-vault) returns hardcoded English strings (`'Invalid OTP URI'`, `'Invalid secret key'`, etc.) and the desktop app surfaces them verbatim via `otpSecretField.error`. Not a blocker for this PR (the strings live in the lib), but worth tracking — non-EN locales will see English here. Consider returning error codes from the lib and translating in the UI, or moving these specific messages through Lingui at the call site.

**Verifications I ran**
- `validateOtpInput` / `parseOtpInput` / `matchLoginRecords` are exported from lib-vault at `3381deda` ✓
- `DESKTOP_2FA_IMPORTS_ENABLED` exported from lib-constants at `4d56e038` ✓
- `Combobox`, `MultiSlotInput.errorMessage`, `PasswordField.rightSlot`/`isGrouped` exist in ui-kit at `ce3ab5de` ✓
- `extractDomainName` has no remaining callers in the desktop repo ✓

PR #508, opened 2026-05-14
Wire vault deletion UI and connect receive-side to action-exchange.

shavtvalishvili:
## Summary
Desktop delete-vault UX wired end-to-end with the receive-side action handler and the new direct-delivery transport (merged in from `device-direct-actions`).

### UI
- Sidebar vault selector and Your-Vaults V2 settings page expose Delete. The V2 confirmation modal validates the master password, runs `deleteVaultLocal`, and falls back to a fresh unprotected "Personal" vault when the last vault is removed (matches `CardCreateMasterPasswordV2`).
- Paired-device count pluralised (`1 other device` vs `N other devices`).

### Receive-side
- `useVaultAccessRevoked` subscribes to `pearpassVaultClient.on('vault-access-revoked', ...)` and runs the local delete via the existing Redux flow.
- Bails out when the vault is no longer in the local list (covers duplicate/self deliveries).
- `master-update` listener fallback route is `navigate('vault', { recordType: 'all' })` (previously v1 welcome+VAULTS which flashed CardVaultSelect during local deletes).

### Send-side
- "Erase from all devices" toggle calls `broadcastDeleteVault(vaultId)` before `deleteVaultLocal`.

### device-direct-actions (now merged)
- Electron IPC proxy exposes the new vault-core primitives: `personalSwarmInit/Close/Send/GetTopic`, `signMessage`, `verifySignature`, `activeVaultFind`, `activeVaultGetWriterKey`, plus the `personal-swarm-envelope` event piped to the renderer.
- Cross-device delivery uses the personal-swarm direct transport with outbox retries (1m for 7d, 1h to 30d, drop); inbound envelopes are ed25519-signed and verified against the sender's master-vault writer pubkey. See lib-vault and lib-vault-core PRs.

### Misc
- Dev hook `__pearpassTriggerAccessRevoked` removed.
- Type declarations updated (src/types/modules.d.ts): ACTION_TYPES, broadcastAction, `@tetherto/pearpass-lib-vault/src/instances`, addDevice signature, new vault-core methods.

## Recordings

https://github.com/user-attachments/assets/9c200f72-9978-4ec5-a60b-7fbc22e9f063

https://github.com/user-attachments/assets/6cbe909d-c640-4e99-ae66-a57a849c1487

https://github.com/user-attachments/assets/0f39fb85-dbc3-4dac-9773-1104fa5e59b7

https://github.com/user-attachments/assets/5bf0896e-0b49-4e2f-9aa6-4888d36fb3e9

https://github.com/user-attachments/assets/60ae5122-6ade-4b2e-97e0-c2684af23eb5

## Depends on
- https://github.com/tetherto/pearpass-lib-vault/pull/48
- https://github.com/tetherto/pearpass-lib-vault-core/pull/59
- https://github.com/tetherto/pearpass-lib-native-messaging-bridge/pull/11


torkinos:
@nickgr2 please review

nickgr2 (commented):
A few items to address before approval — mostly type-declaration consistency and one UX gap. The architecture (master-update + personal-swarm-envelope IPC routing, receive-side hook, V2 dialog with broadcast toggle) reads cleanly.

**Type declaration gap — `vaultOnPersonalSwarmEnvelope`**

`electron/preload.cjs` exposes the new `vaultOnPersonalSwarmEnvelope` binding, and `src/electron/vaultClientProxy.js` calls `api.vaultOnPersonalSwarmEnvelope?.(...)`. But neither `src/types/electron.d.ts` (which lists `vaultOnUpdate` / `vaultOnMasterUpdate`) nor the JSDoc on `createElectronVaultClientProxy` mentions it. Optional chaining masks the gap at runtime, but anything reaching for `window.electronAPI.vaultOnPersonalSwarmEnvelope` from TS will fail to resolve. Mirror what was done for `vaultOnMasterUpdate`.

**Silent failure in fallback Personal-vault creation**

In both `DeleteVaultModalContentV2.onSubmit` (lines ~437–451) and `useVaultAccessRevoked.handleAccessRevoked` (lines ~115–128), if `createVault({ name: t('Personal') })` or the subsequent `addDevice()` throws, only `logger.error` runs. The user is left after a successful local delete with no active vault, no toast, and no modal. The success path emits the `'A new "Personal" vault was created'` toast — the failure path should at least surface a toast (or modal) so the user knows recovery didn't complete and can act.

**Naming inconsistency in `VAULT_METHODS`**

`src/electron/vaultClientProxy.js` adds both `removeVault` and `vaultsRemove` to the same array. `removeVault` also gets a NM bridge handler in `VaultHandlers.js` and an entry in `commandDefinitions.js`; `vaultsRemove` does not. If they refer to the same underlying client method, the duplicate is dead weight. If they're distinct (e.g. `vaultsRemove` removes a master-list entry, `removeVault` deletes the vault entirely), worth a one-line comment so the next reader doesn't have to reverse-engineer it.

**`signMessage` / `verifySignature` exposure — confirm scope**

These primitives are now reachable from the renderer through the IPC proxy. The PR description notes envelope verification happens at the lib layer, but `signMessage` itself is a fairly broad capability — a compromised renderer could request signatures over arbitrary payloads. Is the intent that any caller in the renderer can invoke these, or are they meant to be used only by `pearpassVaultClient` internals? If the latter, worth gating in the proxy so they aren't trivially callable from app code.

**`useUserData().logIn` for password verification**

`onSubmit` calls `await logIn({ password: passwordBuffer })` purely to validate the master password before deletion. Does `logIn` have session side effects (rotating tokens, switching active session, persisting auth state) that aren't desired here? If a non-mutating verification primitive exists, it would be safer; if not, a comment noting the reuse would help future readers.

**Minor**

- `useVaultAccessRevoked.tsx`: `latest.current = { ... }` reassigned on every render outside `useEffect` — works, but conventionally lives in a `useEffect` (or the values are captured directly via the `useCallback` deps). Readability nit.
- `useVaultAccessRevoked.handleAccessRevoked`: when `deleteVaultLocal` throws, the modal still says "Access Removed" while the local files remain. UX-wise correct (access *was* revoked at the source), but worth confirming there's no follow-up retry the user should be able to trigger.
- `Sidebar/VaultSelector.tsx`: removing the `UNSUPPORTED &&` gate exposes Delete unconditionally — matches PR intent, just flagging that any prior reason for the gate is now gone for good.

shavtvalishvili:
@nickgr2, I addressed:
- `vaultOnPersonalSwarmEnvelope` declared in `electron.d.ts`.
- `addDevice` -> `() => Promise<void>` and `setCurrentDeviceName` declared in `types/modules.d.ts`.
- Toast on fallback Personal-vault failure in both `DeleteVaultModalContentV2` and `useVaultAccessRevoked`.
- Comment grouping `removeVault` vs `vaultsRemove` in `VAULT_METHODS`.
- Comment on `logIn` reuse for password verification.
- `latest.current` moved into `useEffect`.
- Tests added: `useVaultAccessRevoked`, `DeleteVaultModalContentV2`, `AccessRemovedModalContent`.

I didn't address:
- `signMessage` / `verifySignature` renderer exposure: the underlying concern is gone since the signing keypair is now separate from the autobase writer keypair (vault-core PR). Renderer-level gating is a separate hardening pass.

nickgr2 (approved):
All review feedback addressed cleanly — approving.

**Resolved:**
- `vaultOnMasterUpdate` / `vaultOnPersonalSwarmEnvelope` typed on `Window['electronAPI']` in `src/types/electron.d.ts`
- Fallback Personal-vault creation failure now surfaces a "Couldn't create a starter vault" toast in both `DeleteVaultModalContentV2.onSubmit` and `useVaultAccessRevoked.handleAccessRevoked`
- `removeVault` vs `vaultsRemove` distinction documented inline in `vaultClientProxy.js` ("disk wipe + master entry" vs "single key from master")
- `logIn` reuse for password verification is now commented at the call site
- `latest.current` reassignment moved inside `useEffect` in `useVaultAccessRevoked.tsx`
- New test coverage for `AccessRemovedModalContent`, `DeleteVaultModalContentV2` (invalid password, broadcast failure toast, fallback success/failure), `useVaultAccessRevoked` (subscribe/unsubscribe, unknown-vault no-op, wipe+switch, fallback success/failure, delete-failure-still-shows-modal), and `nativeMessagingIPCServer` (subscribe, fan-out to clients, client-throw isolation, detach on stop)

**Deferred (acceptable):**
- `signMessage` / `verifySignature` renderer exposure — justified because the signing keypair is now distinct from the autobase writer keypair, so a renderer caller signing arbitrary payloads can't impersonate vault writes
- `AccessRemovedModalContent` shown even when local `deleteVaultLocal` throws — matches the underlying truth that access was revoked at the source; recovery state is best surfaced separately

**Tiny nit, not blocking:**
- `createElectronVaultClientProxy` JSDoc lists `vaultOnMasterUpdate?` on its `api` param but not `vaultOnPersonalSwarmEnvelope?`. The global `electron.d.ts` carries the type and the proxy uses optional chaining, so this is cosmetic — add it next time you touch the file.

giorgikh93 (approved):

PR #510, opened 2026-05-14
feat: enhance import functionality with Bitwarden encrypted files support

ndrkltsk:


ndrkltsk:
@nickgr2 please review this pr

nickgr2 (changes requested):
A few items to address before merge.

**1. Bitwarden encrypted-detection flags non-password-protected exports**

`detectIsEncrypted` (in `src/pages/SettingsViewV2/content/ImportItemsContent/utils.ts`) only checks `parsed?.encrypted === true`. But Bitwarden exports two encrypted variants:

- **Account-restricted** — encrypted with the account key; sets `encrypted: true` only. Cannot be decrypted with a password (re-import only works on the same Bitwarden account).
- **Password-protected** — sets both `encrypted: true` and `passwordProtected: true`. This is what `decryptBitwardenJson` actually supports — its own internal check is `if (!json.encrypted || !json.passwordProtected) throw new Error('File is not password-protected')`.

Right now, an account-restricted export will be detected as encrypted, the UI will prompt for a password, and any input will fail decryption. The outer `catch` in `onImport` then rewrites the underlying `'File is not password-protected'` error to the generic `'Failed to decrypt file. Please check your password and try again.'` — which is misleading.

Suggested fix: parameterize `detectIsEncrypted` by import type so that, for Bitwarden, it also requires `passwordProtected === true`. PearPass-encrypted exports keep the single-flag check.

**2. The outer try/catch in `onImport` swallows actionable error info**

Lines ~293-297 (`index.tsx`):

```ts
} catch {
  throw new Error(
    'Failed to decrypt file. Please check your password and try again.'
  )
}
```

This catches everything — wrong password, malformed file, unsupported KDF, "not password-protected", network/IPC failures from the worklet — and presents all of them as "wrong password." At minimum, distinguish between password mismatch and structural errors (e.g., re-throw non-password errors as-is, or `logger.error` the original before rewriting). Today the same generic message is shown for a corrupted file as for a typo.

**3. Missing test coverage for the new behavior**

`utils.ts` is a new pure function with no test, and the new `decryptBitwardenJson` branch in `onImport` is not exercised in `index.test.tsx`. Per `REVIEW_GUIDELINES.md`, behavior changes need test coverage:

- Unit tests for `detectIsEncrypted` (JSON vs CSV, `encrypted: true` vs missing, malformed JSON, `passwordProtected` once #1 is addressed).
- A flow test that uploads an encrypted Bitwarden JSON, gates on the password screen, and exercises the decrypt path (mocking `decryptBitwardenJson` / `pearpassVaultClient`).

**4. minor: hardcoded English error strings**

`'Password is required for encrypted files'` (line ~280, new) and `'Failed to decrypt file. Please check your password and try again.'` (line ~295) bypass Lingui. These bubble up through `handleImport`'s `error.message` straight into `setErrors` / `setToast`, so users see un-translated English. The second predates this PR, but the new branch keeps adding to the pattern; consider wrapping them in `t(...)` and threading through.

**5. minor: redundant JSON parse**

`detectIsEncrypted` parses the full file just to read one flag, then `onImport` parses it again. For typical exports this is negligible, but for large vaults it's wasted work. Could be folded into the existing parse path or memoised.

ndrkltsk:
@nickgr2 review again please

nickgr2 (approved):
Re-review at bb2df49. All five items from the prior review are addressed:

1. `detectIsEncrypted` (utils.ts) now branches on Bitwarden vs PearPass-encrypted, requiring `passwordProtected === true` for the Bitwarden path. Account-restricted exports correctly skip the password gate.
2. The outer try/catch in `onImport` now logs the original error, distinguishes password-mismatch errors via regex on the message, and re-throws structural errors as-is rather than misattributing them to a bad password.
3. New `utils.test.ts` covers `parseJsonContent` (object/array/primitive/malformed/ArrayBuffer) and `detectIsEncrypted` (CSV vs JSON, Bitwarden variants, PearPass-encrypted, missing flags). New flow tests in `index.test.tsx` cover the encrypted-Bitwarden password gate and the account-restricted direct-parse path.
4. All new error strings wrapped in `t(...)`.
5. JSON is parsed once via `parseJsonContent` in `handleFilesChange`, threaded through `FileInfo.parsedJson`, and reused in `onImport`.

Also verified upstream: `decryptBitwardenJson` is exported at the bumped `pearpass-lib-data-import` SHA, and `decryptBitwardenExport` exists in `pearpass-lib-vault-core` at the bumped SHA.

LGTM.

boris91 (approved):

torkinos (approved):

PR #514, opened 2026-05-18
Implement clearStaleVaultsDir to manage stale vault directories

ndrkltsk:
---
  - https://app.asana.com/0/0/1214093173779876

### DIFF
--- a/electron/clearStaleVaultsDir.cjs
+++ b/electron/clearStaleVaultsDir.cjs
@@ -1,0 +1,48 @@
+const fs = require('fs')
+const path = require('path')
+
+// Recovers from a partial master-password creation: if the app was killed
+// mid-flow on a previous attempt, the `vaults` dir on disk is blind-encrypted
+// with the abandoned password and the next attempt will fail to open it with
+// a new salt. Safe to call only before vaults are opened by the worklet
+// (i.e. before createMasterPassword / login). Refuses to delete when the
+// sibling `vault/` (singular) dir has children — that dir is written only by
+// `initActiveVaultInstance` after a successful `createVault`, so its presence
+// with children is unambiguous evidence the user already has at least one
+// real vault we must not clobber.
+async function clearStaleVaultsDir({ storagePath, logger, fsImpl = fs }) {
+  try {
+    const vaultsDir = path.join(storagePath, 'vaults')
+    const vaultDataDir = path.join(storagePath, 'vault')
+
+    if (!fsImpl.existsSync(vaultsDir)) {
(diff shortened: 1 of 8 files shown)


ndrkltsk:
@nickgr2 please review this pr

torkinos (approved):

nickgr2 (approved):
The recovery logic is sound: `createMasterPassword` writes `vaults/masterEncryption` before `encryptionAdd('masterPassword', …)` in `node_modules/@tetherto/pearpass-lib-vault-core/src/worklet/masterPasswordManager.js`, so being killed in between leaves a `vaults/` dir blind-encrypted with an abandoned `hashedPassword` and a missing `encryption/masterPassword`. Clearing `vaults/` before the retry resolves that. The `vault/` (singular) guard is correct — `initActiveVaultInstance` only writes `vault/${id}` after `createVault` allocates an ID, so the presence of children there is unambiguous evidence of real user data.

A few notes:

- **minor:** the same partial-create bug exists on the v1 path. `src/pages/WelcomePage/CardCreateMasterPassword/index.js` calls `createMasterPassword(passwordBuffer)` with no preceding cleanup. Given v2 is the active design and v1 may be unreachable in practice, this might be intentional — but worth confirming so we don't silently regress v1 users on a downgrade or in the legacy path.

- **minor:** the `eslint-disable-next-line import/order` on the `clipboardCleanup` import in `electron/main.cjs` is a workaround for the new import being inserted out of alphabetical order. Swapping the two lines (clipboardCleanup first, then clearStaleVaultsDir) removes the need for the disable.

- **minor:** `electron/clearStaleVaultsDir.cjs` line 42 logs only `err.message` and drops the rest of the error object. Existing patterns in `main.cjs` (e.g. line 311) pass the full `err` so stack and `code` survive in the log. Since the IPC handler also discards the return value, the message is the only forensic trace if `rmSync` ever fails — losing the stack here will hurt diagnosis on rare partial-create incidents. Suggest passing `err` directly to the logger.

- **question:** the IPC handler in `electron/main.cjs` awaits `clearStaleVaultsDir` but discards the result. Renderer-side `clearStaleVaultsDir` then resolves to `undefined`, so the call site in `CardCreateMasterPasswordV2/index.tsx` can't distinguish "deleted", "refused (had user data)", or "errored". Is fire-and-forget intentional? The downstream `createMasterPassword` will still surface its own error to the user, so functionally it's fine — just flagging that any future telemetry on recovery success/failure will need this to propagate.

Approving on the basis that the core fix is correct and the notes above are non-blocking.

PR #518, opened 2026-05-18
Rewrite native host wrapper on each launch to fix AppImage stale paths.

shavtvalishvili:
### Changes

- Extract wrapper-script generation into `electron/nativeHostWrapper.cjs`; pair-time path in `nativeMessagingSetup.js` now imports it (no behavior change).
- Add `refreshNativeHostWrapperIfPresent()` — atomic rewrite (tmp + rename) when the wrapper exists; no-op when absent.
- Wire `refreshNativeHostWrapper()` into `app.whenReady()` in `electron/main.cjs`: AppImage launches at a fresh `/tmp/.mount_*/` path each run, so paths baked into the wrapper at pair time go stale. Snap is skipped (stable mount path).
- Extract `getRuntimeExecPath()` to dedupe Windows Store execPath logic.
- Add `electron/nativeHostWrapper.test.js` — 9 tests covering per-platform wrapper content, atomic write, and present/absent refresh.

### DIFF
--- a/electron/main.cjs
+++ b/electron/main.cjs
@@ -5,6 +5,7 @@
  * and registers secure IPC handlers so the renderer can use runtime and vault services.
  */
 const fs = require('fs')
+const os = require('os')
 const path = require('path')
 
 const {
@@ -38,6 +39,7 @@ const {
   isFlatpakRuntime,
   isSnapRuntime
 } = require('./flatpak-paths.cjs')
+const { refreshNativeHostWrapperIfPresent } = require('./nativeHostWrapper.cjs')
 const runtimeConfig = require('./runtime-config.cjs')
 const devicePreferences = require('../src/utils/devicePreferences.cjs')
 const {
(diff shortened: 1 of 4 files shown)


boris91 (approved):

torkinos:
@nickgr2 please review

nickgr2 (approved):
Solid fix for the AppImage stale-path issue.

- Atomic write via `tmp-${pid}` + `rename` is the right shape — Chrome can't see a half-written wrapper, and a `rename` after the wrapper is unlinked between `access` and `rename` still does the right thing (just re-creates it).
- Snap skip is correct (stable mount path) and matches the existing `useSnapDirect` branch in `setupNativeMessaging`.
- For the Flatpak case the wrapper content is constant (no bridge/exec paths baked in), so the refresh is a redundant rewrite there — harmless, but you could short-circuit Flatpak too if you want to avoid the disk write at every launch. Not a blocker.
- Renderer's `generateNativeHostExecutable` still writes non-atomically (`writeFile` directly to `executablePath`). If a user pairs while the app is starting and `refreshNativeHostWrapper` runs concurrently, they could race. Outcome is the same content either way, so it's not a correctness issue — flagging for awareness, not asking for a change.
- Test coverage hits the right seams (per-platform content, atomic tmp+chmod+rename, present vs absent). Good.

PR #523, opened 2026-05-19
show scannable QR when viewing recovery phrases

Mysticswiss:
## Summary

When viewing a recovery phrase record in the vault detail panel (v2), show a scannable QR code below the read-only word grid so users can transfer the mnemonic to another device.

- QR is generated only in view mode (`!isCreateOrEdit`); create/edit flows are unchanged.
- Payload is the existing parsed word list joined with spaces (`passphraseWords.join(' ')`), matching what the UI displays.
- Uses `@tetherto/pear-apps-utils-qr` with `margin: 4` and `errorCorrectionLevel: 'L'` for reliable phone and wallet scanning.

## Test plan

- [ ] Open an existing recovery phrase in the vault side panel → QR appears below the words.
- [ ] Scan with a phone camera → decoded text is the space-separated recovery phrase.
- [ ] Scan with a wallet that supports mnemonic import (valid BIP39 12/24-word phrase).
- [ ] Create or edit a recovery phrase → no QR is shown.
- [ ] Wi-Fi QR and Add Device pairing QR still work as before.

## Recordings

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/containers/PassPhrase/PassPhraseV2.tsx
+++ b/src/containers/PassPhrase/PassPhraseV2.tsx
@@ -110,19 +110,18 @@ export const PassPhraseV2 = ({
   }, [value])
 
   useEffect(() => {
-    if (isCreateOrEdit || !value?.trim()) {
+    if (isCreateOrEdit || !passphraseWords.length) {
       setQrSvg('')
       return
     }
-    const words = parsePassphraseText(value)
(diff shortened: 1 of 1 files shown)


Mysticswiss:
Closing to start fresh with a clean PR.

PR #524, opened 2026-05-19
Show scannable QR when viewing recovery phrases

Mysticswiss:
## Summary

Scannable QR on the recovery phrase.

### UI

- Shown inview mode only (`PassPhraseV2`, under the word list).
- QR encodes `passphraseWords.join(' ')` (space-separated).
- Create/edit recovery phrase flows unchanged.

### Image 
Screenshot 2026-05-19 at 4 04 44 PM


torkinos (changes requested):
@Mysticswiss Thanks again for updating the PR and adding more context/screenshots

One suggestion, maybe it would be worth reusing a similar UI/UX approach we already have for the WiFi Record. I think that layout feels more contextual and visually connected to the content around it.

You can find the implementation here:
`/src/containers/WifiPasswordQRCode/WifiPasswordQRCodeV2.tsx`

I’ve attached a screenshot as reference as well 👍

Image

PR #528, opened 2026-05-20
2fa import matching UI

ndolidzee:
### Screenshots/Recordings

https://github.com/user-attachments/assets/fcff9a96-96cb-4696-b9ca-fe20955ff806

### dependencies

---
  - https://app.asana.com/0/0/1214456116873184

### DIFF
--- a/src/pages/SettingsViewV2/content/ImportCodesContent/ScanResultsView.tsx
+++ b/src/pages/SettingsViewV2/content/ImportCodesContent/ScanResultsView.tsx
@@ -7,11 +7,7 @@ import {
   Text,
   useTheme
 } from '@tetherto/pearpass-lib-ui-kit'
-import {
-  Close,
-  ExpandMore,
-  Extension
-} from '@tetherto/pearpass-lib-ui-kit/icons'
+import { LinkOff, ExpandMore, Link } from '@tetherto/pearpass-lib-ui-kit/icons'
 import { useRecords } from '@tetherto/pearpass-lib-vault'
 
 import { RecordItemIcon } from '../../../../components/RecordItemIcon/RecordItemIcon'
@@ -173,7 +169,7 @@ export const ScanResultsView = ({ onImportComplete }: ScanResultsViewProps) => {
                         />
                       ) : (
                         <div style={styles.noMatchIcon}>
-                          <Close
+                          <LinkOff
                             width={16}
                             height={16}
                             color={theme.colors.colorTextSecondary}
@@ -232,7 +228,7 @@ export const ScanResultsView = ({ onImportComplete }: ScanResultsViewProps) => {
                       variant="secondary"
                       size="small"
                       iconBefore={
-                        <Close
+                        <LinkOff
                           width={16}
                           height={16}
                           color={theme.colors.colorTextPrimary}
@@ -253,7 +249,7 @@ export const ScanResultsView = ({ onImportComplete }: ScanResultsViewProps) => {
                       variant="secondary"
                       size="small"
                       iconBefore={
-                        <Extension
+                        <Link
                           width={16}
                           height={16}
                           color={theme.colors.colorTextPrimary}
--- a/src/pages/SettingsViewV2/content/ImportCodesContent/ScanResultsView.tsx
+++ b/src/pages/SettingsViewV2/content/ImportCodesContent/ScanResultsView.tsx
@@ -1,0 +1,304 @@
+import { useCallback, useRef, useState } from 'react'
+
+import {
+  AlertMessage,
+  Button,
+  ListItem,
+  Text,
+  useTheme
+} from '@tetherto/pearpass-lib-ui-kit'
+import {
+  Close,
+  ExpandMore,
+  Extension
+} from '@tetherto/pearpass-lib-ui-kit/icons'
+import { useRecords } from '@tetherto/pearpass-lib-vault'
+
+import { RecordItemIcon } from '../../../../components/RecordItemIcon/RecordItemIcon'
+import { useTranslation } from '../../../../hooks/useTranslation'
+import { getRecordSubtitle } from '../../../../utils/getRecordSubtitle'
+import { CodeMatchDropdown } from './CodeMatchDropdown'
+import { createStyles } from './ScanResultsView.styles'
+
+type ImportedCode = {
+  id: string
+  issuer: string
+  accountName: string
+}
+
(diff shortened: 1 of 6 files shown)


ndolidzee:
@nickgr2 please review

boris91 (approved):

giorgikh93 (approved):

src/pages/SettingsViewV2/content/ImportCodesContent/ScanResultsView.tsx
```diff
@@ -0,0 +1,300 @@
...
+                    </Button>
+                  )}
+                </div>
+              </div>
+            )
+          })}
+        </div>
+      </div>
+
+      <CodeMatchDropdown
```
denislupookov, line 276: hm, can we use DropDown here or Combobox ?

torkinos (changes requested):
@ndolidzee lets please address comment by @denislupookov

denislupookov (approved):

src/pages/SettingsViewV2/content/ImportCodesContent/ScanResultsView.tsx
denislupookov: decided to keep it

ndolidzee:
> @ndolidzee lets please address comment by @denislupookov

@torkinos Me and @denislupookov discussed and we agreed to keep current code, could u please merge ?

torkinos (approved):

nickgr2 (changes requested):
Reviewed against `CLAUDE.md` / `AGENTS.md` / kit source in `node_modules/@tetherto/pearpass-lib-ui-kit`. Scope here is UI scaffolding (mock data, no real wiring), so most comments target the bits that should be obviously placeholder vs. the bits that will ship.

**Blockers / behavioral**

1. `ScanResultsView.tsx` — `MOCK_IMPORTED_CODES` is a hardcoded array of fake codes with no TODO/FIXME marker and no prop to supply real codes. The whole component renders mock content. Either:
   - take `importedCodes` as a prop and feed them in from the scan step (preferred — the data clearly belongs to the scan result, not to this view), or
   - at minimum mark the constant `// TODO: replace with parsed scan output before merge` so a future reader can find it.
   `IMPLEMENTATION_GUIDELINES.md`'s "no placeholders" rule applies here — an unmarked mock literal in a production code path is the failure mode it's meant to catch.

2. `ScanResultsView.tsx:386` — `useRecords({ shouldSkip: true, ... })`. With `shouldSkip: true` the hook never fetches (verified in `useRecords.js:116`), so `filteredRecords` is permanently `[]` and the dropdown will only ever render the "No login records found" empty state. Either drop `shouldSkip`, or gate it on `activeDropdown !== null` so the query runs while the dropdown is open. Right now the matching UI can't actually match anything.

3. `ScanResultsView.tsx:597` — `handleLinkRecord(activeDropdown!.codeId, record)` non-null asserts `activeDropdown` inside the dropdown's `onSelectRecord`. Safe at the moment because the parent only mounts the menu when `activeDropdown !== null`, but the `!` plus the wider "no `any`/no `as`" guidance in `REVIEW_GUIDELINES.md §6` argues for narrowing: capture `activeDropdown.codeId` before passing the callback to `CodeMatchDropdown` (e.g. via a `useMemo` or a derived `selectedCodeId` const) and the assertion goes away.

**Styling discipline**

4. `CodeMatchDropdown.tsx:153` — `borderRadius: 8` hardcoded; everything else in the diff goes through `rawTokens.radius8`. Use the token here too for consistency.

5. `CodeMatchDropdown.tsx:40` — `MENU_BOX_SHADOW` is a hardcoded rgba stack. The kit theme exposes elevation tokens; if there isn't one that matches, raise it with the design-system team rather than inlining the Figma export (this is the pattern `AGENTS.md` is asking for so themes stay consistent in dark mode).

6. `ScanResultsView.styles.ts` — `noMatchIcon: { width: 32, height: 32 }` is hardcoded; `iconSize={32}` is repeated as a magic number in `ScanResultsView.tsx`. If the kit has a size token / sized container variant, use it; otherwise extract a single `ICON_TILE_SIZE` const so the placeholder tile and the real icon can't drift.

**Test IDs**

7. Inconsistent attribute on `Button` — some use `testID` (kit prop, also more conventional in this repo), others use `data-testid` (`ScanResultsView.tsx:551`, `:577`, `:606`, `index.tsx:842`). `react-strict-dom` ends up rendering both as `data-testid` on web, but please pick one (`testID`) and apply it across the new code so E2E selectors stay uniform.

**i18n**

8. `index.tsx:813` — `t(\`Learn more about exporting codes from ${selectedOption.title}\`)` uses template-string interpolation inside `t(...)`, which Lingui can't extract as a stable msgid (each runtime title produces a different key). Use a placeholder: `t('Learn more about exporting codes from {source}', { source: selectedOption.title })`. The sibling `ImportItemsContent/index.tsx:560` has the same pattern — would be good to fix both in one go.

9. `importCodesOptions[0].description` is a longer English paragraph defined at module scope and rendered as `{selectedOption.description}` in the new branch without going through `t()`. The old shape only used `description` as the format-string interpolation argument, which is why it slipped past — now that the field carries real user-facing copy, please wrap it (`description: t('To import your codes...')`) or push the string into the JSX where it's rendered.

**Nits (won't block)**

- `minor:` `paddingBlock: 4`/`paddingInline: 4` in `CodeMatchDropdown.tsx:154` — `rawTokens.spacing4` would match the rest of the file.
- `minor:` `MOCK_IMPORTED_CODES` reuses real-looking emails (`simon.j@gmail.com`, `taylor@gmail.com`, etc.). Swap for clearly fake values (`user1@example.com`) before this lands so screenshots / Storybook don't ship real-looking addresses.
- `minor:` `ScanResultsView.tsx:457` — `divider` is `{}` for the last row to drop the bottom border; cleaner as `index < codeMatches.length - 1 ? styles.tableCellRowDivider : undefined` and let the spread skip it.
- `minor:` The dropdown doesn't auto-focus its search input on open; a quick `inputRef.current?.focus()` after `coords` is set keeps keyboard users productive.
- `minor:` PR description is template-empty — once the scan wiring lands it'd help reviewers to fill in the "Things reviewers should pay attention to" section (e.g. "matching is mocked, real data flow comes in #NNN").

Happy to re-look once the mock list and `shouldSkip` are sorted — the surrounding structure looks fine.

PR #531, opened 2026-05-20
[Desktop] - Google import: QR Decoder for Google import (BarcodeDetector API) + call lib with normalizer + append the OTP

giorgikh93:
### Requirements

Google import - imeplement barcode detector logic
### Changes

- created `decodeQrFromImage` that returns valid uri
### Testing Notes

- tested on custom html page
    ```

    <!DOCTYPE html>
    <html>
    <body>
      <input type="file" id="file" accept="image/*" />
      <pre id="out"></pre>
      <script type="module">
        async function decodeQrFromImage(file) {
          if (typeof BarcodeDetector === 'undefined') {
            throw new Error('QR decoding not supported in this environment')
          }
          const detector = new BarcodeDetector({ formats: ['qr_code'] })
          const bitmap = await createImageBitmap(file)
          const results = await detector.detect(bitmap)
          if (!results.length) throw new Error('No QR code detected in image')
          return results[0].rawValue
        }
    
        document.getElementById('file').addEventListener('change', async (e) => {
          const out = document.getElementById('out')
          try {
            const result = await decodeQrFromImage(e.target.files[0])
            out.textContent = result
          } catch (err) {
            out.textContent = 'Error: ' + err.message
          }
        })
      </script>
    </body>
    </html>

```

### Screenshots/Recordings

https://github.com/user-attachments/assets/ff700c3c-8db5-4b09-b9dc-8dd3f3a52e27

### dependencies

https://github.com/tetherto/pearpass-lib-data-import/pull/15 <-- normalizer

---
  - https://app.asana.com/0/0/1214456116873186


giorgikh93:
@nickgr2 please review

boris91 (dismissed):

boris91 (approved):

torkinos (approved):

nickgr2 (approved):
The new `decodeQrFromImage` utility is clean and well-tested. A few observations:

**Scope vs. title.** The PR title mentions "call lib with normalizer + append the OTP", but the diff only adds the decoder utility (plus a dep bump to pearpass-lib-data-import 5d387bc). The decoder isn't imported anywhere in `src/` yet. If that wiring is intentionally a follow-up, fine — but the title is misleading. Consider trimming the title to just "QR Decoder" so reviewers know what they're approving.

**Platform support.** `BarcodeDetector` ships in Chromium with native backends only on macOS (VisionKit) and Android; on **Linux** it's not implemented, and on **Windows** support depends on Chromium build flags. In an Electron build, this means Linux desktop users will reliably hit the `'QR decoding not supported in this environment'` error. Two thoughts:
- The error message is generic — if the user clicked "Import from Google", they'll have no idea why nothing works. Consider surfacing a more specific "QR import isn't supported on this OS" message at the call site, or guarding the UI entry point altogether on platforms where `BarcodeDetector` is missing.
- If cross-platform support is required (most likely it is for a desktop password manager), a JS fallback like `jsQR` / `zxing-wasm` is worth considering. Just a heads-up — not a blocker for this PR if the call-site PR handles it.

**`createImageBitmap` failures.** A non-image input (or a corrupt image) will reject from `createImageBitmap` with a generic `InvalidStateError` / `DOMException`. That bubbles up unhandled. Worth catching and re-throwing as `'Invalid image file'` (or similar) so the caller can show a friendly message — same shape as the two existing throws.

**Type declaration shape.** `src/types/barcode-detector.d.ts` declares both `class BarcodeDetector` and `var BarcodeDetector` for the same identifier. Standard pattern in `lib.dom.d.ts` is `interface BarcodeDetector` + `declare var BarcodeDetector` (interface for the type, var for the constructor object). Functionally equivalent here under `skipLibCheck`, but the `interface` form matches existing DOM typings and is more conventional. minor.

**Tests.** Solid coverage — happy path, multi-result, both Blob and File inputs, both error paths. The new `rejects` / async `toThrow` matchers in `src/types/jest-globals.d.ts` are a nice addition.

Approving on the utility itself; the platform-support gap is the main thing worth following up on at the call site.

PR #534, opened 2026-05-21
2fa import integration

ndolidzee:
### Screenshots/Recordings

https://github.com/user-attachments/assets/68c4c6bc-896c-464e-bc33-4f49a7e0857a

### dependencies

---
  - https://app.asana.com/0/0/1214930662367847

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6217,7 +6217,7 @@
     "node_modules/@tetherto/pearpass-lib-ui-kit": {
       "name": "pearpass-lib-ui-react-native-components",
       "version": "0.0.40",
-      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#f2d856bb46a8030f51461d06241623bc582d9e37",
+      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#bba64ea96fad376a0f0b87d2729fe2742733d7d7",
       "license": "Apache-2.0",
       "dependencies": {
         "react-strict-dom": "^0.0.55"
--- a/src/pages/SettingsViewV2/content/ImportCodesContent/ScanResultsView.tsx
+++ b/src/pages/SettingsViewV2/content/ImportCodesContent/ScanResultsView.tsx
@@ -1,26 +1,34 @@
-import { useCallback, useRef, useState } from 'react'
+import { useCallback, useMemo, useRef, useState } from 'react'
 
+import type { OTPRecord } from '@tetherto/pearpass-lib-data-import'
 import {
   AlertMessage,
   Button,
   ListItem,
   Text,
+  rawTokens,
   useTheme
 } from '@tetherto/pearpass-lib-ui-kit'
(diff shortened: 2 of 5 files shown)


ndolidzee:
@nickgr2 please review

boris91 (approved):

giorgikh93 (approved):

torkinos (approved):

nickgr2 (changes requested):
This wires the 2FA import flow up to real decoding/matching/persistence, replacing the mocks — nice direction. A few issues block merge as-is.

### Blocking

**1. `normalizeImport` is not exported by the pinned `pearpass-lib-data-import`.**
`index.tsx` imports `normalizeImport` from `@tetherto/pearpass-lib-data-import`, but the version pinned in this PR's `package-lock.json` (0.0.14, `git#c52067f…`) only exports the per-format parsers (`parse1PasswordData`, `parseBitwardenData`, …) — no `normalizeImport`, `detectProvider`, `parseOtpUri`, or `decodeMigrationUri`. This PR's lock change only bumps `pearpass-lib-ui-react-native-components`, so the data-import dep is unchanged from base. At runtime `normalizeImport(decoded)` will throw `normalizeImport is not a function`. The new `OTPRecord` / `NormalizeResult` type declarations in `modules.d.ts` make it type-check, which hides the gap. Either bump `pearpass-lib-data-import` to a version that exports these (and pin it in the lock), or confirm that bump lands first.

**2. `matchLoginRecords` is not exported by the pinned `pearpass-lib-vault`, and isn't declared.**
`ScanResultsView.tsx` imports `matchLoginRecords` from `@tetherto/pearpass-lib-vault`. The pinned vault (`git#1d4f866…`) doesn't export it (its `index.js` has no such symbol), and the vault module block in `src/types/modules.d.ts` was not updated to declare it — so this is both a `tsc` error (TS2305, no exported member) and a runtime `TypeError`. Same root cause as #1: the vault dep isn't bumped here.

### Likely bug

**3. Auto-matching won't populate — stale initial state.**
In `ScanResultsView`, `codeMatches` is seeded via `useState(initialMatches)`, where `initialMatches` is a `useMemo` that calls `matchLoginRecords(..., loginRecords)`. `loginRecords` comes from `useRecords`, which resolves asynchronously — on first render it's `[]`, so no matches are found. When the records arrive, the `useMemo` recomputes, but `useState` only uses its initializer once, so `codeMatches` is never updated and the auto-match silently produces nothing. The `useMemo` dep array `[importedCodes]` also omits `loginRecords`, so it's incomplete regardless. Seed empty and sync via an effect on `[importedCodes, loginRecords]`, or derive matches and merge user overrides, rather than freezing the initial value.

### Minor

- `index.test.tsx` mocks `./ScanResultsView` entirely, so the new matching/create/update logic — the part most likely to break — has no coverage. The upload/scan/error paths are well tested; consider a focused test for `ScanResultsView` (matching, link/unlink, overwrite warning, import dispatch).
- minor: the overwrite indicator uses an inline `<div>` with `gap: ${rawTokens.spacing4}px`; consider moving it into `ScanResultsView.styles.ts` to match the rest of the file's styling pattern.

PR #541, opened 2026-05-28
Add ability to import proton 2fa txt files

ndolidzee:
### Screenshots/Recordings

https://github.com/user-attachments/assets/2794696e-fd4b-417c-b770-5dc9a331aaf8

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6226,15 +6226,13 @@
       }
     },
     "node_modules/@tetherto/pear-apps-lib-feedback": {
-      "name": "pear-apps-lib-feedback",
       "version": "0.0.7",
-      "resolved": "git+ssh://git@github.com/tetherto/pear-apps-lib-feedback.git#f616c1c1f9b9774d17595acf864055674fae64fd",
+      "resolved": "git+ssh://git@github.com/tetherto/pear-apps-lib-feedback.git#6c5df8c71efb303ea980214789147ccf7528a06c",
       "license": "Apache-2.0"
     },
     "node_modules/@tetherto/pear-apps-lib-ui-react-hooks": {
-      "name": "pear-apps-lib-ui-react-hooks",
       "version": "0.0.11",
-      "resolved": "git+ssh://git@github.com/tetherto/pear-apps-lib-ui-react-hooks.git#8273bf9e248bbdaacbfb0b7a16d5b63687990481",
+      "resolved": "git+ssh://git@github.com/tetherto/pear-apps-lib-ui-react-hooks.git#e80d01ebdb02f7011e2f28a52a01447dcec0c645",
       "license": "Apache-2.0",
       "peerDependencies": {
         "react": ">=18.3.1",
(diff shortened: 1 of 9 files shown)


ndolidzee:
@nickgr2 please review

nickgr2 (commented):
A few things to address before merging.

**`ScanResultsView.tsx` changes affect Google Authenticator import too — is this intentional?**

The new record creation lost `username` and the display swapped title/subtitle:

```diff
-  title: entry.code.issuer ?? entry.code.label,
-  otpInput,
-  username: entry.code.label
+  title: entry.code.label,
+  otpInput
```

```diff
-  title={entry.code.issuer ?? entry.code.label}
-  subtitle={entry.code.label}
+  title={entry.code.label}
+  subtitle={entry.code.issuer ? entry.code.issuer : undefined}
```

For a typical Google Authenticator `otpauth://` URI with `issuer=GitHub` and `label=alice@example.com`, this changes:
- Display from "GitHub / alice@example.com" to "alice@example.com / GitHub"
- New login records lose their `username` field

The PR title is scoped to Proton 2FA, but these changes also alter the Google Authenticator path. Could you confirm this is intended (and not just a side-effect of adapting to Proton's data shape)? If intentional, worth calling out in the PR description.

**Fragile decrypt-error detection** (`index.tsx:782`)
```ts
if (/incorrect password/i.test(message)) {
```
This depends on the exact wording from `decryptProtonExport`. If the upstream lib changes the message, the wrong-password UX silently regresses to the generic error alert. Consider exposing an error code/type from `pearpass-lib-vault` instead, or matching on a stable error class.

**`importCodesOptions` and `schema` recreated each render** (`index.tsx`)

Both are defined inside the component body and rebuilt every render. `importCodesOptions` was previously a module-scope `const` — moved inside to use `t()`. Wrap in `useMemo` (with `t` as dep) to keep that hoisted shape, and same for the `Validator.object(...)` schema.

**minor:** button label capitalization is inconsistent — `t('Scan file')` (encrypted path, line ~880) vs `t('Scan File')` (other path, line ~896).

**minor:** `'Google Authenticator'` and `'Proton 2FA'` titles in `importCodesOptions` aren't wrapped in `t()`. Product names typically don't need translation, but the description strings right next to them are translated, so the asymmetry is a bit jarring.

ndolidzee:
@nickgr2 i have addressed your comments, only removing the username and affecting google authenticator is intentional, and we can keep it, assuming that labels would be a better match for title than username.
about decrypt-error detection, we do not plan to change in the lib, and are we using this kind of comparison in some other places, please review one more time

nickgr2:
> only removing the username and affecting google authenticator is intentional

OK, your call — flagging it was about scope (PR title is Proton 2FA but the change touches the Google Authenticator path too). Worth a line in the PR description so anyone reading later doesn't get surprised by the UX shift; not blocking.

> are we using this kind of comparison in some other places

I checked. The two that look similar aren't really comparable:

- `src/services/ipc/MethodRegistry.js:117` — `error.message.includes('DESKTOP_NOT_AUTHENTICATED')`, but that string is a stable error code produced by `createErrorWithCode(SecurityErrorCodes.DESKTOP_NOT_AUTHENTICATED, ...)` — an explicit coded contract.
- `src/pages/SettingsViewV2/content/ReportAProblemContent/index.tsx:108` — `error.message === OFFLINE_TIMEOUT`, where `OFFLINE_TIMEOUT` is a constant the same module throws — both sides controlled locally.

The Proton case is different: it's regex-matching a natural-language string emitted by an upstream dep you don't control. If `pearpass-lib-vault` ever rewords `"Incorrect password"` (e.g. capitalization, "Wrong password", localized), the wrong-password UX silently regresses to the generic `setImportError(message)` path — no test would catch it.

If you don't plan to change the lib, the cheap mitigation is a unit test that imports `decryptProtonExport`, calls it with a bad password against a known-good encrypted blob, and asserts the regex matches — that turns a silent regression into a CI failure when the lib changes. Up to you whether that's worth doing now vs. accepting the risk.

Not blocking either way.

giorgikh93 (approved):

boris91 (approved):

PR #543, opened 2026-05-28
Refactor copy-to-clipboard functionality to conditionally disable bas…

ndolidzee:
…ed on state across multiple forms

### Screenshots/Recordings

https://github.com/user-attachments/assets/4f4502eb-018e-417a-a05d-de544ac21ca6

### DIFF
--- a/src/components/OtpCodeField/index.tsx
+++ b/src/components/OtpCodeField/index.tsx
@@ -33,7 +33,7 @@ export const OtpCodeField = ({
   const { i18n } = useLingui()
   const { theme } = useTheme()
   const styles = createStyles(theme.colors)
-  const { copyToClipboard } = useCopyToClipboard()
+  const { copyToClipboard, isCopyToClipboardDisabled } = useCopyToClipboard()
 
   const { code, timeRemaining, type, period, generateNext, isLoading } = useOtp(
     {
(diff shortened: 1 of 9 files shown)


ndolidzee:
@nickgr2 please review

nickgr2 (changes requested):
The conditional gating itself is sound, but the rollout is incomplete and there's a latent flicker issue worth fixing before this merges.

### Inconsistent rollout — copy affordances still visible elsewhere

The title says "across multiple forms", but several call sites still render a visible Copy button (or copy-control) when the user has disabled clipboard. Clicking does nothing (the hook early-returns false), which is exactly the dead-affordance UX this PR is trying to remove:

- `src/containers/PassPhrase/PassPhrase.tsx` — "Copy recovery phrase" Button (lines 171-187 and 241-273) still renders unconditionally.
- `src/containers/Modal/GeneratePasswordModalContent/GeneratePasswordModalContent.tsx` — the "Copy Password" footer Button (line 248-261) is unconditionally rendered when invoked without onPasswordInsert; the primary action calls copyToClipboard.
- `src/containers/Modal/AddDeviceModalContent/AddDeviceModalContent.tsx` — "Copy vault key" Button (lines 113-125).
- `src/hooks/useConnectExtension.js` — the ExtensionPairingModalContent onCopy hook (line 137) and any control bound to it.

Either gate these too, or call out explicitly why they're excluded.

### Hook flicker — initial state defaults to disabled

`src/hooks/useCopyToClipboard.electron.js` initialises with `useState(true)` and reads localStorage in a `useEffect`. Before this PR, that only affected an early-return inside copyToClipboard and was invisible. With the new conditional rendering, every mount of these forms now flashes the buttons hidden for one render, then reveals them after the effect runs. For the default (clipboard enabled) user, that's a visible glitch on each navigation.

`AppPreferencesContent` already uses the right pattern (lines 51-55):

```ts
useState(() => localStorage.getItem(LOCAL_STORAGE_KEYS.COPY_TO_CLIPBOARD_DISABLED) === 'true')
```

Suggest the hook do the same — synchronous lazy initializer, drop the useEffect.

### Tests not updated

- `src/components/OtpCodeField/index.test.tsx` mocks the hook as `{ copyToClipboard }` only. `isCopyToClipboardDisabled` is undefined, so `!undefined === true` and the button renders by accident — not because the mock reflects reality. Per testing guidance (mocks should mirror real shapes), include `isCopyToClipboardDisabled` in the mock and add at least one case for the disabled branch.
- No coverage for the new conditional render on any of the touched forms.

### Minor

- `IdentityDetailsForm.tsx` includes an unrelated reformat of the `drivingLicensePicture` group object (lines 109-115). Same for `NoteDetailsForm.tsx` line 125 (trailing whitespace). Better squashed into a separate cleanup if intentional, or reverted.

PR #546, opened 2026-05-29
Replace InputField with TextArea for note input

ndolidzee:
### Screenshots/Recordings

https://github.com/user-attachments/assets/06a4427b-fee0-4b6f-a498-0b9be82a07bb

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6318,7 +6318,7 @@
     "node_modules/@tetherto/pearpass-lib-ui-kit": {
       "name": "@tetherto/pearpass-lib-ui-react-native-components",
       "version": "0.0.40",
-      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#8f507d836343f5dfeacf1afba754266aa3f88934",
+      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#43e610d1b3a3c91168b02c8a0a5e8c4269d3e42d",
       "license": "Apache-2.0",
       "dependencies": {
         "react-strict-dom": "^0.0.55"
--- a/src/containers/Modal/CreateOrEditCategoryWrapper/CreateOrEditNoteModalContent/CreateOrEditNoteModalContent.tsx
+++ b/src/containers/Modal/CreateOrEditCategoryWrapper/CreateOrEditNoteModalContent/CreateOrEditNoteModalContent.tsx
@@ -11,6 +11,7 @@ import {
   MultiSlotInput,
   PasswordField,
   Text,
+  TextArea,
   useTheme
 } from '@tetherto/pearpass-lib-ui-kit'
 import { RECORD_TYPES } from '@tetherto/pearpass-lib-vault'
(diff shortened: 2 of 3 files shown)


ndolidzee:
@nickgr2 Please review, changes in package-lock are intentional and required for these changes, it updates lib-ui-kit which includes required changes for TextArea component

nickgr2 (changes requested):
The migration in `CreateOrEditNoteModalContent.tsx` is correct — `TextArea`'s `onChange` is `(value: string) => void`, and the call site was properly updated from `e.target.value` to `value`.

The change in `NoteDetailsForm.tsx` has a behavioral regression, though.

**`NoteDetailsForm.tsx` — read-only copy button is silently lost**

`TextArea` does not support `copyable`, `onCopy`, or any slot. Its full prop surface is `label`, `error`, `disabled`, `onChange`, `testID`, plus passthrough HTML `<textarea>` attributes (`TextArea/TextArea.tsx`). `InputField` rendered an actual copy `Button` when `copyable` was set, invoking `onCopy(value)` (`InputField/InputField.tsx:104-116`).

After this change:
- `copyable` is forwarded via `...rest` onto the underlying `html.textarea` as an unknown attribute — no copy button is rendered.
- `onCopy={copyToClipboard}` binds to the native textarea `copy` DOM event, which passes a `ClipboardEvent` — not the note text — to `copyToClipboard(text)`. That hook expects a string and returns `false` for invalid input (`useCopyToClipboard.electron.js:35-43`), so even manual copy wouldn't work as intended.

Net effect: the copy-to-clipboard button disappears from the read-only note details view, and `copyable`/`onCopy` become dead props.

If keeping `TextArea`, restore the copy affordance explicitly (e.g. a `Button` with the `ContentCopy` icon beside the field, wired to `copyToClipboard(values.note)`), or keep `InputField` for this read-only case where the copy button matters. Either way, drop the now-inert `copyable`/`onCopy` props rather than leaving them as no-ops.

**`isGrouped` removal**

`TextArea` has no `isGrouped` prop, so removing it is necessary — but it also means the field loses the grouped styling `InputField` applied inside `MultiSlotInput` (`InputField.tsx:71-74`). Worth confirming the field still renders correctly within `MultiSlotInput` (borders/rounding) without it.

The `package-lock.json` kit bump and the re-indentation in the attachments map look fine.

ndolidzee:
@nickgr2 After lib-ui-kit update TextArea do support copyable, onCopy, input field renders correctly no need for isGrouped prop. please approve if nothing else

nickgr2:
Re-reviewing against the updated lib-ui-kit now — checking that TextArea's copyable/onCopy support and the dropped isGrouped prop fully resolve the change-request items from the prior review. Will follow up with the verdict.

boris91 (approved):

giorgikh93 (approved):

nickgr2:
This one already landed on `main` (merged ~11:07 UTC), so there's nothing left for me to approve here.

One thing worth a quick sanity-check if it wasn't already verified post-merge: in the kit revision I had available, `TextArea` still only exposed `label`/`error`/`disabled`/`onChange`/`testID` (+ passthrough textarea attrs) — `copyable`/`onCopy` lived on `InputField`/`PasswordField`, not `TextArea`. If the kit bump genuinely added them to `TextArea`, the read-only copy button in `NoteDetailsForm` is fine. If not, those props are inert and the copy affordance won't render. A glance at the note-details view in a build confirms it either way.

PR #550, opened 2026-06-01
Open create or edit modal directly from empty collection view

ndolidzee:
### Screenshots/Recordings

https://github.com/user-attachments/assets/353ebdde-f0af-43ec-bb6f-c95a610deee8

### DIFF
--- a/src/containers/EmptyCollectionView/EmptyCollectionView.tsx
+++ b/src/containers/EmptyCollectionView/EmptyCollectionView.tsx
@@ -3,12 +3,10 @@ import React, { useMemo } from 'react'
 import { Button, Text, Title, useTheme } from '@tetherto/pearpass-lib-ui-kit'
 import { Add, ImportExport } from '@tetherto/pearpass-lib-ui-kit/icons'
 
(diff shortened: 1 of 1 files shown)


ndolidzee:
@nickgr2 please review

boris91 (approved):

nickgr2 (approved):
Routing the empty-collection Add Item button straight to the create/edit modal is a clean improvement.

The new handleAddItem mirrors the established call in AppHeaderContainer (same recordType / selectedFolder / isFavorite shape) and matches the useCreateOrEditRecord type signature. recordType here is always a vault RECORD_TYPE, so it never reaches the password-generator branch inside the hook, and falling back to the add menu only when recordType is 'all' is the right call since no specific type is known there.

minor: the isFavorite: isFavoritesView ? true : undefined branch is effectively dead in this component, since the Add Item button only renders when !isFavoritesView. Harmless and consistent with the shared pattern, so not a blocker.

LGTM.

giorgikh93 (approved):

PR #551, opened 2026-06-01
Fix all folders count

ndolidzee:
### Screenshots/Recordings

https://github.com/user-attachments/assets/4fb4a75b-f361-41a5-8c83-dddcc5f67e1b

### DIFF
--- a/src/containers/Sidebar/Sidebar.tsx
+++ b/src/containers/Sidebar/Sidebar.tsx
@@ -94,6 +94,8 @@ export const Sidebar = () => {
     )
   }, [foldersData])
 
+  const allFoldersCount = customFolders.reduce((sum, f) => sum + f.count, 0)
+
   const favoritesCount =
     (foldersData?.favorites?.records?.length as number | undefined) ?? 0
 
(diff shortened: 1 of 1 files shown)


ndolidzee:
@nickgr2 please review

nickgr2 (changes requested):
The new "All Folders" count and the view it links to now disagree.

`handleAllFoldersClick` (`Sidebar.tsx:118-120`) navigates with no `folder` key. That reaches `MainView.tsx:65-66` as `selectedFolder = undefined`, then `useRecords({ filters: { folder: undefined } })` (`MainView.tsx:77`). In `selectRecords.js:14-16` the folder filter only applies when `filters?.folder || filters?.folder === null` — with `undefined` that guard is false, so **no folder filtering happens** and the view shows every record (of the current type), including records with `folder: null`.

But `allFoldersCount = customFolders.reduce((sum, f) => sum + f.count, 0)` (`Sidebar.tsx:97`) sums only records assigned to a custom folder — `selectFolders.js:31-34` routes folderless records into `noFolder`, never into `customFolders`. So any vault with an unassigned record will display more rows under "All Folders" than the badge shows.

The previous value `recordCounts?.all` matched the click target (both = all records). This change fixes the badge-vs-listed-folders discrepancy but introduces a badge-vs-view one.

Pick one consistent semantic:
- If "All Folders" means "all records" (what the click does today) → keep `recordCounts?.all`.
- If "All Folders" means "records that are in a folder" → `handleAllFoldersClick` must also filter to folder-assigned records, otherwise the count still won't match the list.

As written, the count says "in a folder" while the view says "everything".

Other notes:
- No double-counting risk — `record.folder` is a single scalar (`selectFolders.js:17`, `selectRecords.js:14-16`), so summing per-folder counts is safe.
- The bulk of the diff is pure JSX re-indentation (no elements added/removed/reordered) plus cosmetic line-wraps of the category count and `folderClickRecordType` — fine.
- minor: tsc/eslint were not run in the review environment; the change is type-sound by inspection (`allFoldersCount` is `number`, matching the `count` prop).

PR #562, opened 2026-06-14
feat(desktop): keep PearPass running in system tray when window is closed

### LINKED ISSUE #561: Feature Request: Background mode (Minimize to Tray) to prevent browser extension disconnects
## Description

Currently, when PearPass is closed, the application terminates completely. This breaks the connection with the browser extension, forcing users to either reopen the native app manually or even re-pair the extension from scratch. Since a password manager needs to maintain a stable connection with the browser, it should be able to run silently in the background.

## Proposed Solution

Add an optional setting (disabled by default to preserve current behavior) that allows the app to run in the background when the main window is closed:

   1. Intercept the window close event and call window.hide() instead of terminating the process.
   2. Keep the app alive via the system tray (menu bar / system tray icon) so the connection with the browser extension remains active.
   3. OS Specifics: On macOS, additionally call app.dock.hide() when hidden (and app.dock.show() when restored) to completely remove the icon from the Dock bar while running in the background.
   4. Add a "Quit" option to the tray context menu to allow completely closing the app when needed.

andrey-dru-me1:
Closes https://github.com/tetherto/pearpass-app-desktop/issues/561

Keep PearPass running in the tray when the window is closed, so the browser-extension connection stays alive without a visible window taking up space.

- Intercept window close: hide to tray instead of quitting
- Tray icon with Show/Quit context menu
- Settings toggle persisted via devicePreferences.json
- macOS dock show/hide on tray interaction
- Refactor devicePreferences to generic boolean-key pattern

### Requirements

- The browser extension must remain connected even when no PearPass window is visible.
- Closing the window (red X) must not terminate the app process.
- Users need an explicit "Quit" action in the tray to fully exit.
- The behavior must be opt-in via a setting (off by default).

### Changes

- **electron/main.cjs** — added `Tray` creation, window-close interception (`win.on('close')`), `forceQuit` flag, `backgroundModeEnabled` state, and IPC handlers for show/hide/quitting.
- **electron/preload.cjs** — exposed `backgroundModeEnabled` to the renderer.
- **src/utils/devicePreferences.cjs** — generalized from a single `loggingEnabled` key to a `DEFAULTS` map with `coerce()` for any future boolean preferences. Added `backgroundModeEnabled: false` default.
- **src/utils/devicePreferences.test.js** — updated tests for the new generic key pattern.
- **src/pages/SettingsView/content/AppPreferencesContent/** — added a "Keep running in background" toggle that reads/writes `backgroundModeEnabled` via the preload bridge.
- **src/types/electron.d.ts** — added `backgroundModeEnabled` to the `ElectronAPI` type.
- **src/constants/localStorage.js** — added `BACKGROUND_MODE_ENABLED` constant.
- Snapshot updates (7 files) reflecting the toggle addition.

### Testing Notes

- Manually tested on macOS with the toggle on/off, verifying close-to-tray behavior, tray menu Show/Quit, and dock icon visibility.
- Unit tests for `devicePreferences` updated and passing.
- Verified browser extension stays connected when window is in tray.

- The `forceQuit` flag pattern to distinguish close-vs-quit on macOS — check that `app.quit()` and `app.exit()` are used in the right places.
- The `coerce()` function in devicePreferences — this changes the write path for `loggingEnabled` too. Existing users upgrading will get their `loggingEnabled` re-written as a boolean (already the case) but now via the generic function.
- Windows/Linux tray behavior is untested — `isMac` guards are in place, but dual-platform behavior should be verified.

### Screenshots/Recordings

image
image

### Dependencies

- No external dependencies. Depends on Electron's built-in `Tray` and `Menu` modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,6 @@ test-results
 out/
 
 src/PearPass
+
+# AI-generated plans
+plans/
--- a/electron/main.cjs
+++ b/electron/main.cjs
@@ -12,9 +12,11 @@ const {
   app,
   BrowserWindow,
   ipcMain,
+  Menu,
   nativeImage,
   shell,
-  clipboard
+  clipboard,
+  Tray
 } = require('electron')
 const PearRuntime = require('pear-runtime')
 const getPearRuntimeLegacyStorage = require('pear-runtime-legacy-storage')
@@ -111,6 +113,22 @@ let workletSidecar = null
 
 /** @type {import('@tetherto/pearpass-lib-vault-core').PearpassVaultClient | null} */
 let vaultClient = null
+
+/** @type {import('electron').Tray | null} */
+let tray = null
+
+/**
+ * Persisted background-mode preference (default false).
+ * Read from devicePreferences at startup; updated via IPC at runtime.
+ */
+let backgroundModeEnabled = false
+
+/**
+ * Set to true when the user explicitly chooses "Quit" from the tray menu
+ * (or via Cmd+Q on macOS). Allows the window close event to distinguish
+ * between "user clicked red X" (→ hide to tray) and "user wants to quit".
+ */
+let forceQuit = false
 
 function getExecPath() {
   if (!app.isPackaged) return null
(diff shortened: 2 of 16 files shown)


kdihalas:
@andrey-dru-me1 i verified this on Linux and works just fine

PR #577, opened 2026-08-12
Support pairing multiple browser extensions

nuclon:
## Problem

The app can only be paired with one browser extension. `nm.client.data` stores a single `{ publicKey, pairingState }`, and `nmGetAppIdentity` rejects any second browser with `CLIENT_ALREADY_PAIRED`. Running Chrome and Firefox means unpairing one to use the other.

The transport was never the limit — `setupNativeMessaging` already installs manifests for Chrome, Edge, Chromium, Brave and Firefox, and `sessionStore` already holds many concurrent sessions. The limit was that exactly one client identity could be pinned.

## Approach

Each browser is authorized by its own single-use invitation and tracked in a registry, so any number can be paired at once and revoked individually.

Two new modules under `services/security/`, each owning one document in the encrypted store (which has no enumeration API, so each collection is one JSON document under one key):

- **`pairedClients.js`** → `nm.clients` — the registry: list, add pending, confirm, remove, clear.
- **`pairingInvites.js`** → `nm.pairing.invites` — one-time codes: mint, derive, verify, consume. 10-minute TTL, pruned on read.

`appIdentity.js` sheds its client-pinning half and drops from 528 to 336 lines, keeping only host identity. Both new modules serialize their read-modify-write cycles so two browsers confirming at once can't clobber each other.

Codes are derived with the existing `getPairingCode`, fed a per-invite secret instead of one global secret. They stay deterministic from the stored secret (the modal can be reopened) while single use is enforced by `consumedBy`. The invite is consumed at identity time, not confirm time, closing the window where two browsers race on one code.

## Extension contract change

`nmBeginHandshake` now takes an optional `clientEd25519PublicKeyB64` so the host knows which paired client to sign the transcript for. The transcript keeps its shape and binding — only the resolution of which key goes in changes.

**Backward compatible:** if the parameter is absent and exactly one browser is paired, that one is used, so extensions that haven't shipped the change keep working. With two or more paired, they get a new `AMBIGUOUS_CLIENT` error telling the user to update. `nmFinishHandshake` verifies against the key its session was opened for and needs no extension change.

## Migration

An existing confirmed pairing is adopted into the registry on first read (label `Browser`, date from `nm.identity.creationDate`), so current users keep working across the upgrade. The legacy `nm.client.data` document is left in place for rollback. The locked-state localStorage cache widens from one key to an array, still reading the old key as a fallback.

## UI

Settings lists each paired browser with a user-chosen label, pairing date and a per-row **Unpair** that closes only that browser's sessions. **Add Browser Extension** is always available. A separate destructive control turns the whole integration off — the only path that resets the host identity. A new `AddBrowserModalContent` asks for the label; the pairing modal now shows which browser the code is for and counts down to expiry. All kit components, no new primitives.

Because pairing completes over IPC outside React, `nmConfirmPairing` dispatches a `paired-browsers-changed` event that the new `usePairedBrowsers` hook subscribes to.

## Also in here

- **Deleted `nmResetPairing`** — registered as a handler but absent from `COMMAND_DEFINITIONS`, so unreachable over IPC. Dead code that would unpair every browser if it ever became reachable.
- **Fixed a stray `input.set(publicKey, secret.length)` in `getPairingCode`** — a leftover from before the domain-separation tag, overwriting 23 of the 32 secret bytes with the public key and cutting the secret's contribution to ~9 bytes. Nothing visibly broke since it was deterministic, but it silently weakened the derivation. Safe to fix now because one-time invites replace the old codes.

## Testing

`npx jest` — **761 passing, +50 new tests, 0 new failures.** The 11 remaining snapshot failures (6 suites, all in legacy `lib-react-components`) are identical to `main`; verified by diffing failing test names against a stashed clean tree.

New coverage: registry add/confirm/remove, pending pruning, migration, concurrent writes, corrupt JSON; invite mint/derive/expiry/single-use; two-browser pairing, invite reuse rejection, handshake with and without the client parameter, ambiguity error, per-client signature verification; `closeSessionsForClient` isolation; hook refresh on the pairing event; settings rendering N rows and per-row unpair.

`npx eslint ./src` — 0 errors. `npx tsc --noEmit` — only the pre-existing missing-lingui-locales error, identical on `main`.


nuclon:
Extension-side companion: tetherto/pearpass-app-browser-extension#236 — sends `clientEd25519PublicKeyB64` in `nmBeginHandshake`.

The two can merge in either order. Without the extension change this PR still works for a single paired browser via the fallback; a second browser needs #236 to complete its handshake.

PR #581, opened 2026-09-05
fix: show password mismatch feedback during master password creation

### LINKED ISSUE #579: If the passwords differ, a message must be displayed.
Image

macOS 26.5.2 / PearPass 2.1.0

Hyrudotcom:
Fixes #579.

If the passwords don't match, Continue is disabled but there's no message explaining why. The error was only set on submit, so it never showed up.

This shows “Passwords do not match” under Repeat Password once both fields have a value. The message clears when they match or either field is emptied.

Checked with six local regression tests (not included in this PR), TypeScript, and lint for the changed file. The full suite still has existing snapshot and lint failures. Couldn't verify the desktop app because the build fails on a missing `CLIPBOARD_CLEAR_TIMEOUT` dependency export.


Hyrudotcom:
Could someone review this when they have a chance?

